PR #7, opened 2023-02-14
Add users table

codetheweb:
Closes https://github.com/sfbrigade/project-compass/issues/5 and https://github.com/sfbrigade/project-compass/issues/6.


server/src/db/migrations/1_initial-migrations.sql
```diff
@@ -1,6 +1,15 @@
-CREATE TABLE student (
-  student_id SERIAL PRIMARY KEY,
+CREATE EXTENSION "uuid-ossp";
+
+CREATE TABLE "user" (
+  user_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
```
amantri, line 4: For now, using a UUID is ok. Since we will eventually integrate with Google Workspace for logins and identity, we will replace this with whatever ID Google exposes (whether it be their email address or gaia or something else).

server/src/db/migrations/1_initial-migrations.sql
```diff
@@ -1,6 +1,15 @@
...
   first_name TEXT NOT NULL,
   last_name TEXT NOT NULL,
-  birth_year INT NOT NULL
+  roles TEXT[] NOT NULL CHECK (roles <@ '{"para", "case_manager", "admin"}')
```
amantri, line 7: I think having only one option is better. The current three - admin, case_manager and para are hierarchical in that being an admin can allow you to act as everything, being a CM, you should be able to acts as a para, and the para is the leaf. We can assign folks the highest level role.

Here are some of the responsibilities of the different roles:
admin: misc like creating other roles :)
case_manager: create IEP and tasks, assign paras and schedule
para: collect attempt information for a particular task

server/src/db/migrations/1_initial-migrations.sql
```diff
@@ -1,6 +1,15 @@
...
+);
+
+CREATE TABLE "student" (
+  student_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+  first_name TEXT NOT NULL,
+  last_name TEXT NOT NULL,
+  assigned_case_manager_id UUID REFERENCES "user" (user_id) ON DELETE SET NULL
```
amantri, line 14: Is this a property of the student of the IEP? as in can a student be working with multiple case managers?

server/src/db/migrations/1_initial-migrations.sql
francisli: I'm not sure you want the primary key of the table to be an externally generated id... as you add external login providers, additional columns (unique/indexed as needed) can be added for linking/mapping...

codetheweb: I think we would probably want to have that as a separate column in addition to the `user_id`--ideally primary ID columns have the same type and meaning across all tables in the schema.

server/src/db/migrations/1_initial-migrations.sql
codetheweb: Good question, from what I've seen so far it sounds like students are assigned a case manager but it's worth clarifying.

server/src/db/migrations/1_initial-migrations.sql
codetheweb: I was originally thinking in terms of RBAC when modeling this. I think it's definitely possible that we could introduce new roles in the future that aren't hierarchical, but yeah it's probably simpler to limit to a single role for now.

codetheweb: changed in 0b77c5c

server/src/db/migrations/1_initial-migrations.sql
amantri: I think the direction of case_manager -> student link is more important, than student -> case_manager (i.e. we should probably have the student property in the users table).

server/src/db/migrations/1_initial-migrations.sql
```diff
@@ -1,6 +1,15 @@
...
+
+CREATE TABLE "user" (
+  user_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
   first_name TEXT NOT NULL,
   last_name TEXT NOT NULL,
-  birth_year INT NOT NULL
+  role TEXT NOT NULL CHECK (role IN ('para', 'case_manager', 'admin'))
+);
+
+CREATE TABLE "student" (
```
amantri, line 10: Let's also add an email field here, which can be used to uniquely identify students that share the same names

server/src/db/migrations/1_initial-migrations.sql
amantri: Good point - having our own IDs makes sense.

amantri (approved):

server/src/db/migrations/1_initial-migrations.sql
codetheweb: One case manager has many students though, right?

amantri: Yes, a CM will have multiple students. One of the first things we will have to build is the CM dashboard, where they can  add/remove students and edit their IEPs.

PR #16, opened 2023-02-22
Rename tables to be plural

amantri:


codetheweb:
I'm not going to die on this hill, but I much prefer singular names in SQL. Here's some good reasons why: https://stackoverflow.com/a/5841297

amantri:
Fair enough - will drop this PR.

PR #17, opened 2023-02-23
Fetch student list from server and render them on the client

amantri:


client/pages/students.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, [])
+
+  if (isLoading) return <p>Loading...</p>
+  if (!data) return <p>No profile data</p>
+
+  return (
+    <>
+      <title>Student list</title>
+
+      <div>
```
amantri, line 33: The current table is quite ugly, but I didn't want to spend too much time fighting the css quite yet :)

client/pages/students.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export default function Students() {
+  const [data, setData] = useState(null)
+  const [isLoading, setLoading] = useState(false)
+
+  useEffect(() => {
+    setLoading(true)
+    // TODO(amantri): Get backend url from env
+    fetch('http://localhost:8080/students/list')
```
amantri, line 18: Is fetch the right way to get make the api call, or should I be using axios or something else?

codetheweb: there's a ton of data fetching libraries for React
generally I wouldn't use `fetch` directly because you then need to deal with error & loading states yourself (and miss out on a lot of goodies from other libraries)

[React Query](https://tanstack.com/query/latest/docs/react/overview) is a very popular and easy to use data fetching library.

[SWR](https://swr.vercel.app/) is another moderately popular one that doesn't provide quite as much functionality, but maybe a little simpler.

client/pages/students.tsx
```diff
@@ -0,0 +1,55 @@
+import { useState, useEffect } from 'react'
+
+// TODO(amantri): get this from the server code, preferably auto-generated
+interface Student {
+  student_id: string;
+  first_name: string;
+  last_name: string;
+  assigned_case_manager: string;
+}
```
codetheweb, line 9: another option is [trpc](https://trpc.io/) which would eliminate needing to duplicate types like this, but not sure how many layers of abstraction the team wants to learn at once

server/src/app.ts
```diff
@@ -1,13 +1,22 @@
+import cors from "cors";
 import express from "express";
 import { withDb, withEnv } from "./lib";
 import { withMorgan } from "./lib/middleware/with-morgan";
 import { healthRouter } from "./routes";
 import { studentsRouter } from "./routes";
 
+// TODO(amantri): Get client port from env. Do we add this only on dev env?
```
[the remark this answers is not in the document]
amantri: I try not to use wildcards for issues like this in case we forget to remove it before deploying :)

It's better to have things break during development than to introduce security issues.

client/pages/students.tsx
amantri: Thanks for the pointers. Switched to SWR

client/pages/students.tsx
amantri: Yeah - I'd be inclined to using trpc (or even grpc*) for this. Are the Zapatos generated types an option here?

* we will need to define the json schema for our db somewhere. Protobufs (with json marshallers) are a pretty solid choice here IMHO. jscon-schema looks to be way too verbose.

** aside: there is also grpc-js: https://www.npmjs.com/package/@grpc/grpc-js

PR #19, opened 2023-02-27
Run 'npm install' on init to ensure that modules and deps are up to date.

amantri:


francisli (approved):
that should do it...!

amantri:
Thanks for the suggestion. I tested it out, and it works.

amantri:
Thanks for the suggestion. I tested it out, and it works.

PR #20, opened 2023-03-01
TRPC integration

codetheweb:
Notice that we don't explicitly define the types anywhere, but everything is typed correctly in the client:

Screenshot 2023-02-28 at 7 28 18 PM

Demo of new pages, showing an example of querying and mutations (starting on `http://localhost:3000/students`):

https://user-images.githubusercontent.com/7410405/225818308-83747ebc-446a-43d4-b37c-2fb6ffb6a7fc.mov


client/tsconfig.json
```diff
@@ -16,9 +16,10 @@
     "incremental": true,
     "baseUrl": ".",
     "paths": {
-      "@/*": ["./*"]
+      "@/*": ["./*"],
+      "~/*": ["../server/src/*"]
```
codetheweb, line 20: this is kinda hacky, didn't take the time to set up a base tsconfig.json properly

amantri:
This is very cool and magical!

If I understand correctly, the student type is being defined by the output of the db query in trpc.ts. Is the generated type in a file somewhere that we can look at?

codetheweb:
> If I understand correctly, the student type is being defined by the output of the db query in trpc.ts. Is the generated type in a file somewhere that we can look at?

The database query builder is typed with Zapatos and the generated types are in `db/zapatos/schema.d.ts`. If you're asking whether there's persistent generated types for the trpc query itself, no--it's all implicitly typed on the fly.

server/src/routes/trpc.ts
```diff
@@ -0,0 +1,57 @@
...
+    .query(async (req) => {
+      const { student_id } = req.input;
+
+      const result = await req.ctx.db
+        .selectFrom("student")
+        .where("student_id", "=", student_id)
+        .selectAll()
+        .executeTakeFirst();
+
+      return result;
```
amantri, line 28: It would be good to be explicit about the core types in our system like `student`. We can define interfaces for those (either explicitly or via protobufs ;) and import those on the client.

Adding a TODO for this is ok for now.

client/tsconfig.json
amantri: Can you please add this as a comment in the code here. A pointer to an existing article (or new article/doc in our wiki) can be included to keep the comment brief.

server/src/routes/trpc.ts
```diff
@@ -0,0 +1,57 @@
...
+    }),
+
+  getAllStudents: t.procedure.query(async (req) => {
+    const result = await req.ctx.db.selectFrom("student").selectAll().execute();
+
+    return result;
+  }),
+
+  createStudent: t.procedure
```
amantri, line 37: Please add a TODO to check for duplicates

amantri, line 37: Related to the type comment above - what is the best practice for define the interface of request and return objects in trpc? In grpc, there is typically a Request and Response message that defines the input/output of a method. This method has only two input params so this is ok, but for some of the more complicated flows, we can easily have a dozen or more input params that we need the client to send.

client/pages/students/[student_id]/view.tsx
```diff
@@ -0,0 +1,27 @@
+import { trpc } from "@/lib/trpc";
+import { useRouter } from "next/router";
+
+const ViewStudentPage = () => {
```
Comment (amantri) [inline client/pages/students/[student_id]/view.tsx:4]: Should this file be called index.ts so a url like `frontend_server/students/<someid>` can use NextJS' file based routing?

client/references.d.ts
```diff
@@ -0,0 +1 @@
+import "../server/src/db/zapatos/schema";
```
amantri, line 1: Re filename: `references` isn't a great name as it is ambiguous about what is being referred to here - consider `schema`?

amantri (approved):
Looks great!

codetheweb:
Realized last night I forgot to add tests. 😅

Added in 87bf9e4.

server/src/routes/trpc.ts
codetheweb: should we treat names as unique?

client/pages/students/[student_id]/view.tsx
Comment (codetheweb) [reply client/pages/students/[student_id]/view.tsx:4]: yeah, doesn't matter too much--was just an example, but can move

server/src/routes/trpc.ts
codetheweb: the request interface is defined by the `.input(...)` builder function with the Zod schema

response interfaces aren't currently explicitly defined but can be done similarly with `.output(...)`

client/references.d.ts
codetheweb: `references.d.ts` feels slightly better to me over `schema.d.ts` since our client doesn't actually care about our database schema, it just happens to be that Zapatos declares its types in a `.d.ts` which can't be implicitly resolved like everything else--but fine changing if you think `schema.d.ts` is a better fit :)

server/src/routes/trpc.ts
amantri: Names aren't unique, but we can treat emails as unique.

client/references.d.ts
amantri: We have two schemas - storage and communication. This is the communication schema.

server/src/routes/trpc/students.ts
```diff
@@ -0,0 +1,40 @@
+import { z } from "zod";
+import { t } from "./context";
+
+// todo: define .output() schemas for all procedures
```
amantri, line 4: nit: I recommend using this convention for TODOs: https://google.github.io/styleguide/pyguide.html#312-todo-comments

client/references.d.ts
codetheweb: sorry, I think there might be some confusion--`zapatos/schema.d.ts` are types generated directly from our Postgres database, not types that we manually define

server/src/routes/trpc.ts
codetheweb: got it, wasn't sure if we were going to track student emails

oos for this pr so I'll add a todo for now

PR #24, opened 2023-03-06
Only lint TS & JS in pre-commit hook

codetheweb:


amantri (approved):

package.json
```diff
@@ -22,7 +22,7 @@
     "prettier": "2.8.3"
   },
   "lint-staged": {
-    "**/*": [
+    "**/*.{js,jsx,ts,tsx}": [
       "prettier --write --ignore-unknown",
```
amantri, line 26: We should run prettier on all files.

PR #25, opened 2023-03-06
Run prettier on all files

amantri:


codetheweb (commented):
Screenshot 2023-03-07 at 6 32 49 PM

I think this has a possible race condition

amantri:
Yeah, I looked at the documentation, but couldn't figure out where to pass in the "--concurrent" flag or add the "concurrent=false" setting. Any suggestions as to how to turn off the concurrent execution?

codetheweb:
I would recommend adding a new rule for `*.json` and *.md` instead of modifying the existing one.

Otherwise, `./husky/pre-commit` is where the flag would be added.

PR #40, opened 2023-03-30
Consolidate servers

codetheweb:
Apologies for the large diff, kinda inevitable.


tsconfig.json
```diff
@@ -14,9 +14,10 @@
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
-    "baseUrl": ".",
+    "baseUrl": "./src",
     "paths": {
       "@/*": ["./*"],
```
codetheweb, line 19: I'd like to maybe rename this to
```suggestion
      "client/*": ["./*"],
```

for consistency but that can be a future PR

.env.example
```diff
@@ -0,0 +1,5 @@
+# Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
+DATABASE_URL=postgres://postgres:@localhost:5432/compass
+
+GOOGLE_CLIENT_ID=foo
```
amantri, line 4: Are these needed to the get the app working?

codetheweb: not currently, good point

I'll leave a note

amantri:
Does a client/server or frontend/backend directory structure make sense here to help clarify what is being executed where?

For example, here is one possible layout:
- frontend (existing src folder)
  - pages
  - public
  - styles
  - components (isn't there yet, but will be there soon)
 
- backend (existing api folder, to be made sibling of frontend)
  - db
  - lib
  - routers (is `trpc` a better name to avoid confusion with Nextjs' page based routing?)
  - scripts

One benefit of having such a delineation is to help navigate the project code base, and also have some rules of thumb like code in the backend directory should have unit tests etc.

codetheweb:
Would moving `scripts` to `src/api/scripts` and/or maybe renaming `src/api -> src/backend` work?

Unfortunately Next.js requires the client structure to either be `src/pages` or `pages/`.

amantri:
> Would moving `scripts` to `src/api/scripts` and/or maybe renaming `src/api -> src/backend` work?
> 
> Unfortunately Next.js requires the client structure to either be `src/pages` or `pages/`.

Oh ok. Given Nextjs' requirement, your proposal above src sounds good.

amantri (approved):

PR #42, opened 2023-04-10
Add persistent session-based auth adapter

codetheweb:
This configures NextAuth to use a custom, persisted adapter instead of the default stateless JWT one.

This lays the groundwork for authenticating incoming requests and attaching them to a user.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,9 @@
 # Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
 DATABASE_URL=postgres://postgres:@localhost:5432/compass
 
-NEXTAUTH_SECRET=  # Generate using `openssl rand -base64 32`
+# This should be a long random string in production
+# One way to generate a suitable value is using `openssl rand -base64 32`
+# NEXTAUTH_SECRET=
 
 # Only needed if you want to use sign in with Google
 # GOOGLE_CLIENT_ID=
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 # Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
 DATABASE_URL=postgres://postgres:@localhost:5432/compass
 
-NEXTAUTH_SECRET=example-secret # Generate using `openssl rand -base64 32`
+NEXTAUTH_SECRET=  # Generate using `openssl rand -base64 32`
 
 # Only needed if you want to use sign in with Google
 # GOOGLE_CLIENT_ID=
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 # Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
 DATABASE_URL=postgres://postgres:@localhost:5432/compass
 
-NEXTAUTH_SECRET=example-secret # this should be a long random string in production
+NEXTAUTH_SECRET=example-secret # Generate using `openssl rand -base64 32`
 
 # Only needed if you want to use sign in with Google
 # GOOGLE_CLIENT_ID=
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,8 @@
 # Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
 DATABASE_URL=postgres://postgres:@localhost:5432/compass
 
+NEXTAUTH_SECRET=example-secret # this should be a long random string in production
+
 # Only needed if you want to use sign in with Google
 # GOOGLE_CLIENT_ID=
 # GOOGLE_CLIENT_SECRET=
--- a/src/backend/auth/adapter.ts
+++ b/src/backend/auth/adapter.ts
@@ -24,6 +24,13 @@ const mapStoredSessionToAdapterSession = (
   expires: session.expires_at,
 });
 
+/**
+ * Factory function to create a custom Auth.js adapter.
+ * At time of writing, there isn't an official adapter that works with our setup.
+ * Having a custom adapter also gives us more control over the User model.
+ *
+ * See https://next-auth.js.org/tutorials/creating-a-database-adapter
+ */
 export const createPersistedAuthAdapter = (
   db: KyselyDatabaseInstance
 ): Adapter => ({
@@ -39,6 +46,7 @@ export const createPersistedAuthAdapter = (
         last_name,
         email: user.email,
         email_verified_at: user.emailVerified,
+        // todo: this should be pulled from an invite or something else instead of defaulting to a para
         role: "para",
         image_url: user.image,
       })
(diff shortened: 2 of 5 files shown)


codetheweb:
Still not sure who I should tag for review 😅 

Maybe we can add a CODEOWNERS file as this becomes more active and have auto-assigned round robin reviewers.

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -4,7 +4,32 @@ CREATE TABLE "user" (
...
   last_name TEXT NOT NULL,
-  role TEXT NOT NULL CHECK (role IN ('para', 'case_manager', 'admin'))
+  role TEXT NOT NULL CHECK (role IN ('para', 'case_manager', 'admin')),
+  email TEXT UNIQUE NOT NULL,
+  email_verified_at TIMESTAMPTZ,
+  image_url TEXT
+);
+
+CREATE TABLE "account" (
+  account_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
```
amantri, line 14: What is an account_id modeling?

codetheweb: sorry, are you asking what the table models or what the specific `account_id` column does?

the table represents the connection between users and third-party SSO providers like Google OAuth

it could technically be one table if you're only ever going to support one SSO provider but I think this is a little cleaner (and what their official adapters do)

amantri: Thanks - I was asking about what both, the account table, and the account_id field were modeling.

Re user and account table: I imagine that if someone logs in with different SSO providers or even different IDs from the same SSO provider (e.g. different Gmail addresses), they are essentially different users in our system. For now, a simple way to do ACLs is by domain - for e.g. we allow sfusd.edu CM accounts to work with other sfusd.edu para accounts and so on.

Is it a reasonable assumption that each Oauth identity is a separate "user" on Compass? If yes, do we need separate user and account tables?
If not, how will we manage user and account tables?

codetheweb: no, NextAuth considers an email to be a user's unique identity

if:
1. someone signed in with Google OAuth using the email `foo@gmail.com`
2. they signed in again with GitHub OAuth, and their GitHub account has the same email `foo@gmail.com`

there would be two rows in `account` (one for Google and one for GitHub) and a single row in `user`

we don't manage any of this ourselves, it's handled by NextAuth through our custom adapter

src/backend/auth/adapter.ts
```diff
@@ -0,0 +1,195 @@
...
+  sessionToken: session.session_token,
+  userId: session.user_id,
+  expires: session.expires_at,
+});
+
+export const createPersistedAuthAdapter = (
+  db: KyselyDatabaseInstance
+): Adapter => ({
+  async createUser(user) {
+    const [first_name, last_name] = user.name?.split(" ") ?? [
```
amantri, line 31: nit: splitting on space doesn't always work, for e.g. when someone's name has more than two parts.

src/backend/auth/adapter.ts
```diff
@@ -0,0 +1,195 @@
+import { Adapter, AdapterSession, AdapterUser } from "next-auth/adapters";
```
amantri, line 1: OOC, why did you choose not to use the existing adapters, for say Prisma or Sequelize?

Also, add a link to the interface being implemented in the comments in this file: https://next-auth.js.org/tutorials/creating-a-database-adapter

src/backend/auth/adapter.ts
```diff
@@ -0,0 +1,195 @@
...
+      "",
+    ];
+    const createdUser = await db
+      .insertInto("user")
+      .values({
+        first_name,
+        last_name,
+        email: user.email,
+        email_verified_at: user.emailVerified,
+        role: "para",
```
amantri, line 42: Add a TODO to get the role from the url or some other means.

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -4,7 +4,32 @@ CREATE TABLE "user" (
...
   first_name TEXT NOT NULL,
   last_name TEXT NOT NULL,
-  role TEXT NOT NULL CHECK (role IN ('para', 'case_manager', 'admin'))
+  role TEXT NOT NULL CHECK (role IN ('para', 'case_manager', 'admin')),
+  email TEXT UNIQUE NOT NULL,
+  email_verified_at TIMESTAMPTZ,
+  image_url TEXT
+);
+
+CREATE TABLE "account" (
```
amantri, line 13: Please add a comment mentioning the table's use, its relation to the user table, and maybe also indicate this is managed by the NextAuth adapter

.env.example
```diff
@@ -1,7 +1,7 @@
 # Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
 DATABASE_URL=postgres://postgres:@localhost:5432/compass
 
-NEXTAUTH_SECRET=example-secret # Generate using `openssl rand -base64 32`
+NEXTAUTH_SECRET=  # Generate using `openssl rand -base64 32`
```
codetheweb, line 4: I think a default value here is helpful for dev environments to suppress the warning?

src/backend/auth/adapter.ts
codetheweb: yeah, this is more of a best-effort

ideally users will be able to edit their name somewhere in the app

src/backend/auth/adapter.ts
codetheweb: please see new comment further down in this file

.env.example
amantri: The main concern I have with a default value is that it might get pushed to prod accidentally. IMHO, for security related things, it is better to have an error and slight inconvenience for the developer against the risk of accidental prod push.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,14 +1,25 @@
...
+const handler: NextApiHandler = (req, res) => {
+  const { db } = createContext({
+    req,
+    res,
+  });
 
-    // ...add more providers here
-  ],
-});
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
Comment (amantri) [inline src/pages/api/auth/[...nextauth].ts:13]: Interesting - I don't get any error on this line for `npm run lint`

amantri (approved):

.env.example
codetheweb: I disagree but not a big deal

src/pages/api/auth/[...nextauth].ts
Comment (codetheweb) [reply src/pages/api/auth/[...nextauth].ts:13]: Screenshot 2023-04-13 at 9 57 21 PM

hmm, did you `npm i`?

the error makes sense to me since the function is typed as any:

Screenshot 2023-04-13 at 9 58 08 PM

PR #49, opened 2023-04-14
Run `npm i` as part of `npm run dev`

amantri:


amantri:
An alternative to this change is for me to change my invocation command to:
$ npm i && npm run dev

codetheweb:
Interesting idea. I'm personally quite strongly against this, because
- it's not conventional (I've never seen an open source JS/TS project that does this)
- `npm run dev` is only 1 possible command, what about `npm run lint` or `npm run db:reset`?
- developers working in JS/TS codebases are generally expected to know that they need to run `npm i` when `package.json` changes, usually if you haven't pulled in a while from `main` or you're checking out someone else's branch

I would be more open to adding `npm i` as a Git hook via huksy when the branch changes, but again that would be extremely unconventional.

amantri:
> Interesting idea. I'm personally quite strongly against this, because
> 
> * it's not conventional (I've never seen an open source JS/TS project that does this)
> * `npm run dev` is only 1 possible command, what about `npm run lint` or `npm run db:reset`?
> * developers working in JS/TS codebases are generally expected to know that they need to run `npm i` when `package.json` changes, usually if you haven't pulled in a while from `main` or you're checking out someone else's branch
> 
> I would be more open to adding `npm i` as a Git hook via huksy when the branch changes, but again that would be extremely unconventional.

If knowing when to run `npm install` is part of the base expectation for a JS/TS engineer, then I will drop this PR. The main motivation for this PR stemmed from observing that several folks on Compass ran into issues due to not running `npm install`, but we can tackle that with better education as well.

PR #51, opened 2023-04-15
update file structure (include dashboards with styling) and add initial para list management 

hannahmcg:
#28


src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -40,3 +40,10 @@ CREATE TABLE "student" (
   last_name TEXT NOT NULL,
   assigned_case_manager_id UUID REFERENCES "user" (user_id) ON DELETE SET NULL
 );
+
+CREATE TABLE "para" (
```
codetheweb, line 44: maybe we want to use the `user` table instead of introducing a new one?

next.config.js
```diff
@@ -14,4 +14,6 @@ const nextConfig = {
   },
 };
 
-module.exports = nextConfig;
+module.exports = {
+  nextConfig,
```
codetheweb, line 18: was the linter or something complaining about this?

src/backend/db/migrations/1_initial-migrations.sql
hannahmcg: definitely! should be correctly updated in the most recent commit now

hannahmcg: @codetheweb ^

.gitignore
```diff
@@ -22,3 +22,5 @@ yarn-error.log*
 
 .next
 next-env.d.ts
+.env.local
```
amantri, line 25: This can be removed - it is covered by line 14 above

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -1,5 +1,6 @@
 CREATE EXTENSION "uuid-ossp";
 
+-- We think there should be a field for para's tables to hold CM id's
```
amantri, line 3: We do need a new table, say "case_manager", that contains the links for para_ids and student_ids. Something like 

// Not sure about postgres syntax, so here is the pseudo code:
Table case_manager (
  para_id - array of user_ids that are a foreign key into "user"
  student_id - array of student_ids that are a foreign key into "student"
  // ... other fields in the future
)

tessathornberry: Do you think this is a task for this PR, or can we make a new PR to handle this?

src/backend/routers/paras.ts
```diff
@@ -0,0 +1,59 @@
...
+      // todo: add a unique constraint to prevent duplicate paras
+      const result = await req.ctx.db
+        .insertInto("user")
+        .values({ first_name, last_name, email, role })
+        .returningAll()
+        .execute();
+      return result;
+    }),
+
+  deletePara: procedure
```
amantri, line 47: "Delete" should really be archive - i.e. remove the link from the CM->para, but not actually delete the para row in the database. If we actually delete the para, all the data that they might have collected now has a dangling user_id.

I would recommend removing this function or changing it to archive.

In general, I don't think we should ever delete (remove database rows) from the frontend - if we want to delete/cleanup things, that can be handled by a "gardener" job that aims to maintain the referential integrity (this is a common pattern in NoSQL world, in SQL world, maybe we can create those constraints in the data model, so we might not need this job)

src/backend/routers/paras.ts
```diff
@@ -0,0 +1,59 @@
...
+      return result;
+    }),
+
+  getAllParas: procedure.query(async (req) => {
+    const result = await req.ctx.db.selectFrom("user").selectAll().execute();
+
+    return result;
+  }),
+
+  createPara: procedure
```
amantri, line 26: This can be shared with creating a CM flow as well.

Right now, that is is adapter.ts -> createdUser, which can likely be refactored into users.ts (being introduced in PR 55)

src/backend/db/migrations/1_initial-migrations.sql
amantri: Ideally, that table would be created in either this PR or in Hieu's PR to manage student roster, whichever happens to come first :). Those two PRs will add the para_id[] and student_id[] fields respectively in the table above. 

Another option is to create the table in a separate PR, and rebase this PR on that, but in general it is useful for a PR to be fully self-contained for a particular piece of functionality.

Without that table, this PR then becomes a way to create users in the database (and currently delete but see comment below) without associating them with a CM, which is orthogonal to where we want the CM dashboard to evolve.

src/components/CaseManagerUI/paras.tsx
```diff
@@ -0,0 +1,60 @@
+import { trpc } from "client/lib/trpc";
+import Link from "next/link";
+import React from "react";
+
```
amantri, line 4: Re parent directory naming: in general, prefer snake_case for directories, e.g. "case_manager". Suffixes like "UI" can be dropped as  this is already in the components directory. 

The actual components themselves can PascalCased as recommended by AirBnb's style guide: https://github.com/airbnb/javascript/tree/master/react#naming

src/backend/routers/paras.ts
tessathornberry: We could do this with a link/unlink or CM ID property if the para has a CM_id involved. It may also be worthwhile considering that a CM's dashboard will only display THEIR paras and the tasks they have assigned to their Para; A CM should ONLY see the tasks they have assigned to THEIR Paras and not a Para's entire task list, if it possibly includes tasks from other CM's. This will require a lot more restructuring, including of the schema, to implement.

src/backend/routers/paras.ts
amantri: This function can be removed. An entry in the user table is created when someone logs in.

src/components/case_manager/AllParas.tsx
```diff
@@ -0,0 +1,78 @@
+import { trpc } from "client/lib/trpc";
+import Link from "next/link";
+import React from "react";
+import styles from "../../styles/Home.module.css";
+
+const AllParasPage = () => {
```
amantri, line 6: Please add some documentation about the purpose of this component, and also any props it expects/users.

src/components/case_manager/AllParas.tsx
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  if (isLoading) {
+    return <div>Loading...</div>;
+  }
+
+  return (
+    <div>
+      <div
+        style={{
```
amantri, line 32: Please extract the css into its own module file.

A TODO is fine for now

src/styles/Home.module.css
```diff
@@ -72,6 +74,41 @@
   background-color: #3023b8;
 }
 
+/* CM DASHBOARD */
+.cmContainer {
```
amantri, line 78: This should ideally be in its own module file

amantri (approved):
Please also update the PR title and description to reflect what is included in this PR

src/styles/Home.module.css
hannahmcg: I agree, I just put it here for now since this was a minimal styling attempt - but I'll make sure to do so in a future PR with more styling changes! @amantri

src/backend/routers/paras.ts
hannahmcg: I see what you mean - just for clarity's sake, would that imply that we have an "invite para" button in place of a create para button for when a case manager is looking to add a para to their roster? And/or would we just have an "add para" button that searches through the existing list of users? @amantri

amantri: Ah - I see what you mean. Yeah, if its a new para, we will want to create a row for them in the user table when the CM adds them. If they already are in the system, we can reuse their user_id. I think you probably need a call to getUserByEmail in auth/adapter.ts here.

hannahmcg: After thinking it through a bit more I think I will make a separate PR to change the functionality of creating a para so that we can consolidate the paras.ts with the users.ts first (which I think Tessa is going to handle in a separate PR) @amantri

PR #57, opened 2023-04-20
Use svg logo instead of png

amantri:
Also consolidate the public directories to the top level as expected by NextJS


keshao728 (approved):
Approve!

PR #59, opened 2023-04-24
Add Student

hieungo89:
Added email field for when adding a new student and added the route to return to student list when in individual student's profile.


src/backend/db/zapatos/.eslintrc.json
amantri, line 1: Undo the changes done by prettier in this file. This file is not ignored by prettier (after you sync to head)

src/backend/routers/students.ts
```diff
@@ -25,14 +25,20 @@ export const studentProcedures = {
...
+        first_name: z.string(),
+        last_name: z.string(),
+        email: z.string(),
+      })
+    )
     .mutation(async (req) => {
-      const { first_name, last_name } = req.input;
+      const { first_name, last_name, email } = req.input;
 
       // todo: add a unique constraint to prevent duplicate students
```
amantri, line 38: This todo can now be removed

src/backend/routers/students.test.ts
```diff
@@ -25,6 +26,7 @@ test("getAllStudents", async (t) => {
     .values({
       first_name: "Foo",
       last_name: "Bar",
+      email: "foo.bar@email.com",
```
amantri, line 29: It would be good to add a test case that tries to insert a duplicate email into the db.

src/backend/db/zapatos/.eslintrc.json
hieungo89: Didn't realize that this has been changed. Will revert it.

src/backend/routers/students.test.ts
hieungo89: I added the new test case for duplicate emails. Please take a look and also let me know if the naming is accurate to describe the test.

amantri (approved):

PR #66, opened 2023-05-01
added roboto font to global style and a redirect-to-cm-dashboard link…

tessathornberry:
added roboto font to global style (Design team choice) and a redirect-to-cm-dashboard link on the landing page for ease of movement around the current parts of the app on the browser.

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -30,7 +30,7 @@ const Home: NextPage = () => {
                 Welcome {me?.first_name} {me?.last_name}
               </h1>
               {JSON.stringify(me)}
-              <Link href={`/cmDashboard`}>
+              <Link href="/cmDashboard">
                 <p>CM Dashboard</p>
               </Link>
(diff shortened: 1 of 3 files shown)


src/pages/index.tsx
```diff
@@ -29,6 +30,9 @@ const Home: NextPage = () => {
                 Welcome {me?.first_name} {me?.last_name}
               </h1>
               {JSON.stringify(me)}
+              <Link href={`/cmDashboard`}>
```
amantri, line 33: OOC, why use backticks here, instead of say <Link href="/cmDashboard">? There is no variable substitution being done here.

amantri (approved):

src/pages/index.tsx
tessathornberry: @amantri Thank you so much! I fixed it - I must have learned that from stack overflow, but now I have corrected the href to the proper format per the next.js docs.

PR #77, opened 2023-05-14
Add admin framework

codetheweb:
Closes https://github.com/sfbrigade/compass/issues/73.

Demo:

https://github.com/sfbrigade/compass/assets/7410405/a640ef77-13c7-46dc-bda3-c727805b2667

### DIFF
--- a/src/backend/auth/adapter.ts
+++ b/src/backend/auth/adapter.ts
@@ -35,6 +35,15 @@ export const createPersistedAuthAdapter = (
   db: KyselyDatabaseInstance
 ): Adapter => ({
   async createUser(user) {
+    const numOfUsers = await db
+      .selectFrom("user")
+      .select((qb) => qb.fn.count("user_id").as("count"))
+      .executeTakeFirstOrThrow();
+
+    // First created user is an admin
+    // todo: this should be pulled from an invite or something else instead of defaulting to a para - currently devs signing in are being assigned as paras
+    const role = Number(numOfUsers.count) === 0 ? "admin" : "para";
+
     const [first_name, last_name] = user.name?.split(" ") ?? [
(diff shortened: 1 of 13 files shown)


src/backend/auth/adapter.ts
```diff
@@ -35,6 +35,15 @@ export const createPersistedAuthAdapter = (
   db: KyselyDatabaseInstance
 ): Adapter => ({
   async createUser(user) {
+    const numOfUsers = await db
+      .selectFrom("user")
+      .select((qb) => qb.fn.count("user_id").as("count"))
+      .executeTakeFirstOrThrow();
+
+    // First created user is an admin
```
amantri, line 43: Nicely done :)

src/backend/auth/adapter.ts
```diff
@@ -35,6 +35,15 @@ export const createPersistedAuthAdapter = (
...
+    // todo: this should be pulled from an invite or something else instead of defaulting to a para - currently devs signing in are being assigned as paras
```
amantri, line 44: I think we will want an admin panel where admins can designate others as admins.

We will likely not allow upgrading a user to admin from emails alone - we want email invites to be shareable.

amantri (approved):

src/backend/auth/adapter.ts
codetheweb: > I think we will want an admin panel where admins can designate others as admins.

Yep, that can be in a future PR.

> We will likely not allow upgrading a user to admin from emails alone - we want email invites to be shareable.

Not quite sure what you mean by this?

src/backend/auth/adapter.ts
tessathornberry: This works great, Max!

PR #78, opened 2023-05-17
Student iep Container

patrickwchoi:
Very simple first PR for an empty container for student's IEPs. Please let me know if there's anything that could've been done better or neater.

I defined a type for Student in my tsx file, but I feel like there was a better way of doing that since the Student type seems to be defined in the backend with zapatos, but I wasn't sure how to use this. 

Thanks


.env.example
```diff
@@ -1,6 +1,6 @@
 # Assumes the Postgres server was started using the Compose file in docker/postgres/docker-compose.yml
 DATABASE_URL=postgres://postgres:@localhost:5432/compass
-
+# generated using `openssl rand -base64 32`
```
codetheweb, line 3: I think this is already specified below?

src/pages/students/[student_id]/studentIEP.tsx
```diff
@@ -0,0 +1,22 @@
+type Student = {
+  //do we need this here?
+  student_id: string;
+  first_name: string;
+  last_name: string;
+  email: string;
+  assigned_case_manager_id: string | null;
+};
```
Comment (codetheweb) [inline src/pages/students/[student_id]/studentIEP.tsx:8]: there is a way to derive this from trpc types, but since this component only uses the first and last name I'd recommend adjusting the props to take `{ first_name: string, last_name: string }` or similar instead of `{ student: Student }`

.env.example
patrickwchoi: oops didnt realize I edited this. my bad

src/pages/students/[student_id]/studentIEP.tsx
Comment (patrickwchoi) [reply src/pages/students/[student_id]/studentIEP.tsx:8]: I figured since this component was meant to display the IEP when we make it, it'd make sense to pass in more than just the student's name because we want to access the IEP information later. but I can still change this for now.

src/pages/students/[student_id]/studentIEP.tsx
Comment (amantri) [inline src/pages/students/[student_id]/studentIEP.tsx:1]: This file can be named iep.tsx

src/pages/students/[student_id]/iep.tsx
```diff
@@ -0,0 +1,15 @@
+interface StudentIEPProps {
+  first_name: string | undefined;
+  last_name: string | undefined;
```
Comment (codetheweb) [inline src/pages/students/[student_id]/iep.tsx:3]: 🐧 convention in TS for possible undefined/missing values is to do
```suggestion
  first_name?: string;
  last_name?: string;
```

codetheweb (approved):

Issue #62, opened 2023-04-27
Make relational tables/schema and archive flow for students and CMs [ENG]


tessathornberry:
In order to make an archive property for students, we will need relational tables to describe CM's relation to students, using the UUID's for both parties:

This may change slightly based upon if/how we restructure the CM and Para to both be Users.
- [x] Backend Make schema/table for CMs-to-students
- [x] Backend: make a "create student" action in the front end flow through this sequence of queries in the back end(while carrying CM uuid as part of the req object or otherwise having access to it in the backend):
    - [x] Backend:  Query for the student (by email) in the Student table 
    - [x] Backend query: If they do not exist, add student from query body to their primary table
    - [x] Backend: Either way, retrieve uuid of student (while carrying CM uuid)
- [x] ~Backend: Put CM's uuid and student's uuid in CM-to-student relational table (Do we need student names here also?)~

- [x] Frontend: Render student lists as the response to queries of the CM-to-student table (CM's uuid in that specific relational table)
- [x] Front-end: Make button for archiving student

- [x] Back-end: have CM-adding-student flow in. routers/students.ts:
* Is student email already in student table?
    * If no, add student
    * If yes, is assigned_case_manager_id **null?**
        * if no, send error message ('there is already a CM assigned to this student')
        * if yes (it's null), add case current manager's id to assigned_case_manager_id property

![image](https://user-images.githubusercontent.com/105292208/234747397-89cc0887-e4c8-42a3-8366-b1ca6136f014.png)
![image](https://user-images.githubusercontent.com/105292208/234747451-ad224673-a07e-4c9d-a9a1-a357bd9de957.png)

amantri:
For the student change, we need to update the student table, and for the paras, we need to introduce a new case_manager_para relation table. Since these two are different, it might be good to break this out into two separate issues.

amantri:
Thanks Tessa!

tessathornberry:
OK, I updated the comment per thte current work:

- [ ] Back-end: have CM-adding-student flow in. routers/students.ts:
* If email is already in student table
    * If no, add student
    * If yes, check if assigned_case_manager_id is null
        * if no, send error message ('there is already a CM assigned to this student')
        * if yes (it's null), add case manager's id to assigned_case_manager_id property

hieungo89:
I made 5 case scenarios when adding students:
- Student not in database
- Student in database with no case manager
- Student in database with different case manager
- Student already added by same case manager
[...]

PR #79, opened 2023-05-17
Archiving Student

hieungo89:
This PR is associated with issue #62.

Completed feature to allow case managers to successfully:

Adding Student
- Add new students that are not in the database
- Add new students that are in the database, but not associated with any case manager
- When adding new students, alert if:
  - student is assigned to another case manager
  - student is already added to the classroom
  - internal error

Archive Student
- Remove the case manager id from the student table
- included with a confirmation step before archiving

**NOTE:**
In order to add a student with no issues, the user needs to sign in via Google since the database requires the CM id. This id is only possible upon signing in.


src/backend/routers/students.test.ts
```diff
@@ -1,15 +1,17 @@
 import test from "ava";
 import { getTestServer } from "backend/tests";
+import { string } from "zod";
```
codetheweb, line 3:
```suggestion
```

src/backend/routers/students.ts
```diff
@@ -17,30 +16,85 @@ export const studentProcedures = {
 
       return result;
     }),
+  //* this function gets students for the current logged in case manager
```
codetheweb, line 19: I would either introduce a new procedure or rename this to make the functionality more clear, maybe something like `getMyAssignedStudents()`?

KCCPMG: I might suggest either getMyStudents() (there's no other context in which anyone can 'have' students, right?) or getAssignedStudents() (if the convention is to not give a perspective to the procedures). If this procedure's name is changed, should the `AllStudents.tsx` component also have its name changed?

tessathornberry: Ok, I changed the name for that function to "getMyStudents", to be in alignment with the "getMe" function name and the future "getMyStaff"/"getMyParas" name (Because we wouldn't say "getAssignedStaff" since I think the dynamic between Paras and Case Managers is different than that of students and Case Managers.

src/pages/students/[student_id]/index.tsx
```diff
@@ -11,12 +15,22 @@ const ViewStudentPage = () => {
     { enabled: Boolean(student_id) }
   );
 
+  const { mutate } = trpc.archiveStudent.useMutation();
+
+  const archiveStudent = async () => {
+    mutate({
+      student_id: student?.student_id || "",
```
Comment (codetheweb) [inline src/pages/students/[student_id]/index.tsx:22]: we should probably do a `if (!student) { return }` check above instead of defaulting to an empty string, since that would throw an error (TS will then realize that `student` must be defined in the code that follows)

src/backend/routers/students.ts
```diff
@@ -17,30 +16,85 @@ export const studentProcedures = {
...
+  getMyStudents: procedure
+    .input(z.object({ assigned_case_manager_id: z.string().uuid() }))
+    .query(async (req) => {
+      const { assigned_case_manager_id } = req.input;
```
codetheweb, line 23: going by the function name, I would expect that this would take 0 parameters and instead look at the user ID from the authentication attached to the request--try `req.auth...`

src/backend/routers/students.ts
```diff
@@ -17,30 +16,85 @@ export const studentProcedures = {
...
-      const result = await req.ctx.db
-        .insertInto("student")
-        .values({ first_name, last_name, email })
-        .returningAll()
-        .execute();
+      const inDatabase = await req.ctx.db
+        .selectFrom("student")
+        .selectAll()
+        .where("email", "=", email)
+        .executeTakeFirst();
```
codetheweb, line 51: [`ON CONFLICT`](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT) can be used in Postgres to perform upserts and reduce the custom logic here.

with kysely I think it's something like `.onConflict(qb => qb.column("email"))...`

package.json
```diff
@@ -30,6 +30,7 @@
     "pg": "^8.9.0",
     "pg-connection-string": "^2.5.0",
     "postgres-migrations": "^5.3.0",
+    "prettier": "^2.8.8",
```
amantri, line 33: OOC: Why move this from dev to prod dependencies?

hieungo89: I'm not quite sure why it's in there. I can move it back to dev dependencies.

src/pages/students/[student_id]/index.tsx
Comment (hieungo89) [reply src/pages/students/[student_id]/index.tsx]: I modified this as you said.

src/backend/routers/students.ts
hieungo89: Completed. Still have some type-error left behind. Not sure how to go around fixing it.

src/backend/routers/students.ts
hieungo89: I'm not sure how the `.onConflict` method can reduce the custom logic when resulting in different error messages. I looked into the other approaches, but still haven't found a good solution.

codetheweb: thinking through this a bit more, it seems like there's potentially two separate methods here: one to create a student and one to update / assign a case manager?

hieungo89: Yes. The only way I can think of for getting a student's `assigned_case_manager_id` is by looking up that student in the query. So I can keep the `const inDatabase` and utilize `.onConflict` to reduce the custom logic, or is there a better way to do this?

src/backend/routers/students.ts
hieungo89: Restarted my vs code and it's working now. I guess I needed to update my vs code.

src/backend/routers/students.ts
```diff
@@ -17,9 +16,15 @@ export const studentProcedures = {
 
       return result;
     }),
+  //* this function gets students for the current logged in case manager
```
codetheweb, line 19: 🐧 the convention in JS/TS is to comment functions like this (called JSDoc):
Screenshot 2023-05-25 at 9 09 10 PM

if that format is followed, then the comment is shown when hovering over the function in frontend routes:

Screenshot 2023-05-25 at 9 09 21 PM

src/components/case_manager/AllStudents.tsx
```diff
@@ -5,10 +5,16 @@ import styles from "../../styles/Dashboard.module.css";
...
+  const user_id = me?.user_id || ""; //temp solution due to typing errors
+
   const utils = trpc.useContext();
-  const { data: students, isLoading } = trpc.getAllStudents.useQuery();
+  const { data: students, isLoading } = trpc.getMyStudents.useQuery();
+
   const { mutate } = trpc.createStudent.useMutation({
-    onSuccess: () => utils.getAllStudents.invalidate(),
+    onSuccess: () => utils.getMyStudents.invalidate(),
+    //in future PR, we could change this to notification instead of browser alert [tessa]
```
codetheweb, line 16: made an issue for this: https://github.com/sfbrigade/compass/issues/84

src/components/case_manager/AllStudents.tsx
```diff
@@ -5,10 +5,16 @@ import styles from "../../styles/Dashboard.module.css";
 import PersonCreationForm from "./PersonCreationForm";
 
 const AllStudentsPage = () => {
+  const { data: me } = trpc.getMe.useQuery(); //adds user_id
+  const user_id = me?.user_id || ""; //temp solution due to typing errors
```
codetheweb, line 9: we should bail early in `handleSubmit()` if `!me` instead of providing this default, since the mutation will fail if `assigned_case_manager_id = ""`

src/backend/routers/students.ts
```diff
@@ -30,17 +35,64 @@ export const studentProcedures = {
...
+      } else if (
+        inDatabase?.assigned_case_manager_id === assigned_case_manager_id
+      ) {
+        throw new Error("You already added this student to your classroom.");
+      } else {
+        throw new Error("Error adding student. This is a database error.");
+      }
     }),
+
+  archiveStudent: procedure
```
codetheweb, line 81: 🐧 maybe 
```suggestion
  unassignStudent: procedure
```
?

src/backend/routers/students.test.ts
```diff
@@ -72,9 +76,10 @@ test("doNotAddDuplicateEmails", async (t) => {
       first_name: "Foos",
       last_name: "Bar",
       email: "foo.bar@email.com",
+      assigned_case_manager_id: seed.para.user_id,
```
codetheweb, line 79: 🐧  somewhat oos but maybe we should move student creation into `src/backend/tests/seed.ts`

src/backend/routers/students.ts
codetheweb: the way I would do it is have a single SQL statement built with something like:

```ts
await req.ctx.db
  .insertInto("student")
  .values({ first_name, last_name, email, assigned_case_manager_id })
  .onConflict(qb => qb.column("email").doUpdateSet({
    assigned_case_manager_id
  }).where("student.assigned_case_manager_id", "=", null))
  .execute()
```

this will:
- attempt to insert into `student`
- if there's an existing row with the same `email` value, update the `assigned_case_manager_id` **only if** it's currently null

codetheweb: I would also probably rename to something like `createStudentOrAssignManager`

src/components/case_manager/AllStudents.tsx
hieungo89: I think about removing lines 8 and 9 entirely and using `const { userId } = req.ctx.auth;` in the backend.

src/backend/routers/students.ts
hieungo89: That's why I kept getting the alert, `assigned_case_manager_id is ambiguous`!

I made the change, but it doesn't seem to add the student in when the `assigned_case_manager_id = null`. Apparently, Postgres sees the evaluation as [false](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-is-null/). The terminology used in Postgres is `IS NULL`, but I couldn't find its equivalent in tRPC.

src/backend/routers/students.ts
hieungo89: Noted. This might be useful to be added on the Wiki @tessathornberry.

src/backend/routers/students.ts
hieungo89: Actually, I figured it out. I'll push the updated codes.

src/backend/routers/students.test.ts
hieungo89: I'm not sure what you mean by moving to that folder. Most of the tests require the use of `seed`. Also, the `users.test.ts` also contains `seed`.

codetheweb: sorry, meant we could add a test student in the seed function instead of re-creating in every test

but oos, not needed for this pr

codetheweb (approved):
nice!

PR #88, opened 2023-05-31
paraflow: automatic email invitation to para

BrettEastman:
- Add Nodemailer module for sending email, located at src/config/nodemailer.ts
- Add logic to src/backend/routers/paras.ts to send email at the same time para email first entered into db by cm


src/backend/routers/paras.ts
```diff
@@ -34,13 +35,23 @@ export const paraProcedures = {
     )
     .mutation(async (req) => {
       const { first_name, last_name, email, role } = req.input;
+      // In the SCHEMA, email has the "unique" constraint TEXT UNIQUE NOT NULL, which will prevent duplicate paras from being inputted
+
+      // this is where Nodemailer will send an email to the email address of the para inputted by the cm
+      await transporter.sendMail({
```
codetheweb, line 41: might make more sense to send the email after inserting to our database, don't want to send an email if the insert fails for some reason

src/components/case_manager/AllParas.tsx
```diff
@@ -21,6 +21,7 @@ const AllParasPage = () => {
       email: data.get("email") as string,
       role: "para",
     });
+    // email invitation here? Or should be in the backend as part of mutate?
```
codetheweb, line 24:
```suggestion
```

src/config/nodemailer.ts
```diff
@@ -0,0 +1,18 @@
+import { createTransport } from "nodemailer";
+
+const email: string | undefined = process.env.EMAIL;
+const pass: string | undefined = process.env.EMAIL_PASS;
```
codetheweb, line 4: these should be added to `src/backend/lib/types/env.ts`

also would be good to document them in `.env.example`

src/components/case_manager/AllParas.tsx
codetheweb, line 24:
```suggestion
```

codetheweb:
also would be good to add a test, could be as simple as adding an event handler on the node mailer object and updating a global state variable with sent messages when `process.env.NODE_ENV === "test"`

src/backend/routers/paras.ts
```diff
@@ -34,13 +35,23 @@ export const paraProcedures = {
...
+        ...mailOptions,
+        to: email,
+        subject: "Para-professional email confirmation",
+        text: "Email confirmation",
+        html: "<h1>Email confirmation</h1><p>Please confirm your email by going to the following link: <a>no link yet</a></p>",
```
amantri, line 46: Please add a TODO to insert link later

src/config/nodemailer.ts
amantri, line 1: Not sure this is config - maybe lib/nodemailer.ts would be a better home for this file.

src/backend/routers/paras.ts
BrettEastman: Hi Max, in paras.ts, I moved transporter.sendMail to after .execute. I'm not sure if that is what you meant. Please let me know.

codetheweb: looks good!

src/backend/lib/nodemailer.ts
```diff
@@ -0,0 +1,18 @@
...
+    user: email,
+    pass,
+  },
+});
+
+// this just means we are sending an email from our own email for testing purposes. The "to" email will be changed in src/backend/routers/paras to the para email address entered into the db by the cm.
+export const mailOptions = {
+  from: email,
+  to: email,
+};
```
codetheweb, line 18: is this needed since the only parameter that's used here is `from`?

src/backend/routers/paras.ts
```diff
@@ -34,13 +35,25 @@ export const paraProcedures = {
...
+      // In the SCHEMA, email has the "unique" constraint TEXT UNIQUE NOT NULL, which will prevent duplicate paras from being inputted
 
-      // todo: add a unique constraint to prevent duplicate paras
       const result = await req.ctx.db
         .insertInto("user")
         .values({ first_name, last_name, email, role })
         .returningAll()
         .execute();
+
+      // this is where Nodemailer will send an email to the email address of the para inputted by the cm
```
codetheweb, line 46:
```suggestion
```

src/backend/routers/paras.ts
```diff
@@ -34,13 +35,25 @@ export const paraProcedures = {
...
+      await transporter.sendMail({
+        ...mailOptions,
+        to: email,
+        subject: "Para-professional email confirmation",
+        text: "Email confirmation",
+        html: "<h1>Email confirmation</h1><p>Please confirm your email by going to the following link: <a>no link yet</a></p>",
+      });
+      // to do here: when site is deployed, add url to html above
+      // to do elsewhere: add "email_verified_at" timestamp when para first signs in with their email address (entered into db by cm)
```
codetheweb, line 55: could be wrong but I think this is already done by our auth adapter

src/backend/routers/paras.ts
```diff
@@ -34,13 +35,25 @@ export const paraProcedures = {
     )
     .mutation(async (req) => {
       const { first_name, last_name, email, role } = req.input;
+      // In the SCHEMA, email has the "unique" constraint TEXT UNIQUE NOT NULL, which will prevent duplicate paras from being inputted
```
codetheweb, line 38:
```suggestion
```

src/backend/lib/nodemailer.ts
BrettEastman: Fixed

codetheweb (approved):
would be nice to get a test in but otherwise looks good!

PR #96, opened 2023-06-02
Switch to actions/checkout@v2

amantri:
actions/checkout@v1 uses Node 12, which is deprecated. 
actions/checkout@v2 uses Node 16 which is the recommended replacement

See https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12


.github/workflows/pr.yml
```diff
@@ -1,12 +1,12 @@
 name: PR
 on: [pull_request]
 jobs:
-  type-test:
-    name: Type test
+  type-check:
+    name: type-check
```
codetheweb, line 5: 🐧 this is what appears in the UI which is why I prefer something more human readable

codetheweb (approved):
did you mean to open this against main?

.github/workflows/pr.yml
amantri: right - that is why I changed them to align them with the `npm run` target names. If a test fails, there is one less layer or translation from the web UI to the CLI command to test locally.

amantri:
> did you mean to open this against main?

Yeah - I created this branch in the UI, so didn't realize until later that it was against `terraform` instead of `main`. Anyway, I was hoping that #81 and #87 would be merged into main soon making that rebasing moot :)

I have more branches that depend on those to send out for review - I prefer smaller changes that build on each other than mega PRs.

Issue #72, opened 2023-05-05
CM flow: Make relational tables/schema and archive flow for paras/users and CMs [ENG]


tessathornberry:
In order to make an archive property for paras, we will need relational tables to describe CM's relation to paras, using the UUID's for both parties:

ALSO make function getAllParas --> getMyParas

This may change slightly based upon if/how we restructure the CM and Para to both be Users.
- [x] Backend Make schema/table for CMs-to-paras
- [x] Backend: make a "create para" action in the front end flow through this sequence of queries in the back end(while carrying CM uuid as part of the req object or otherwise having access to it in the backend):
    - [x] Backend:  Query for the para (by email) in the Para table 
    - [x] Backend query: If they do not exist, add para from query body to their primary table
    - [x] Backend: Either way, retrieve uuid of para (while carrying CM uuid)
- [x] Backend: Put CM's uuid and para uuid in CM-to-para relational table

- [x] Frontend: Render para lists as the response to queries of the CM-to-para table (CM's uuid in that specific relational table)
- [x] Front-end: Make button on Para's individual page for archiving para

![image](https://user-images.githubusercontent.com/105292208/234747397-89cc0887-e4c8-42a3-8366-b1ca6136f014.png)
![image](https://user-images.githubusercontent.com/105292208/234747451-ad224673-a07e-4c9d-a9a1-a357bd9de957.png)

tessathornberry:
next steps - 
[ ] - separate adding para to relational table after adding to user table so multiple cm's can add the same para

PR #105, opened 2023-06-08
Para Flow - create para, assign to cm, archive para, para tests

hieungo89:
Addressing issue #72.

- [x] createPara
    - [x] Added relational table for cm and para
    - [x] Modified to create a para and add para and add cm_id and para_id to the relational table
    - [x] Changed name to createParaAndAssignCaseManager
- [x] Create an unassignPara backend method to delete the row that associates cm to para when archiving.
- [x] para.test.ts - completed tests for all possible conditions.

completed with @KCCPMG, @tessathornberry, and @patrickwchoi.


hieungo89:
I'm not quite sure what's happening with the tests. It was working fine before I merge the `main` branch into `archivePara` branch.

codetheweb:
I suspect your `package-lock.json` got corrupted somehow since there's a 4k line diff but you didn't change `package.json`:

Screenshot 2023-06-08 at 8 44 16 PM

Try `git checkout main package-lock.json` and commit to reset your changes to `package-lock.json`.

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -11,7 +11,11 @@ CREATE TABLE "user" (
 );
 
 -- table for CM's para list will go here
-
+CREATE TABLE "cm_to_para" (
```
codetheweb, line 14:
```suggestion
CREATE TABLE "paras_assigned_to_case_manager" (
```

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -11,7 +11,11 @@ CREATE TABLE "user" (
...
+  relation_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
```
codetheweb, line 15:
```suggestion
```

(for many-to-many tables, we don't need an extra ID--we can make a composite primary key from the two IDs below)

src/backend/routers/paras.test.ts
```diff
@@ -2,19 +2,119 @@ import test from "ava";
...
-  t.is(para.user_id, user_id);
+  const { trpc, seed } = await getTestServer(t, { authenticateAs: "admin" });
+
+  const para = await trpc.getParaById.query({ user_id: seed.para.user_id });
+  t.is(para.user_id, seed.para.user_id);
+});
+
+test("getMyParas", async (t) => {
+  const { trpc, db, seed } = await getTestServer(t, {
+    authenticateAs: "admin",
```
amantri, line 13: This doesn't really need 'admin' privileges - consider changing this to 'para' (to be changed to 'staff' soon)

src/backend/routers/paras.ts
```diff
@@ -18,30 +17,61 @@ export const paraProcedures = {
...
+      .innerJoin("cm_to_para", "user.user_id", "cm_to_para.para_id")
+      .where("cm_to_para.case_manager_id", "=", userId)
+      .selectAll()
+      .execute();
 
     return result;
   }),
 
-  createPara: procedure
+  createParaAndAssignCaseManager: protectedProcedure
```
amantri, line 33: In general, "and" should almost never be part of a function name - it's a code smell - it suggests that the function is doing too much. Consider structuring this as two functions - `createPara` and `assignParaToCaseManager`. This way, the two functions can be called together or separately as needed (we will need to call them separately if the CM is adding a para that already exists).

You can make `createPara` idempotent, where successive calls of the function for the same args are no-ops.

src/pages/paras/[user_id]/index.tsx
```diff
@@ -12,19 +14,63 @@ const ViewParaPage = () => {
...
       </p>
+      <p>
+        <b>Para Email:</b> {para?.email}
+      </p>
+      <button
+        className={`${styles.signIn} ${styles.bold}`}
+        onClick={() => setArchive(true)}
+      >
+        {" "}
+        Archive Para
```
Comment (amantri) [inline src/pages/paras/[user_id]/index.tsx:47]: s/Archive/Remove/g

what does "archiving" a para mean? "Remove para" is the natural opposite of "Add para"

Comment (patrickwchoi) [reply src/pages/paras/[user_id]/index.tsx:47]: We're not deleting the para, only unassigning the para with the CM here. In our backend we name the function "unassignPara". I think the reason we put "archive para" in the frontend was maybe because saying Remove Para is misleading because we aren't deleting the para. Should we say "Unassign Para" instead?

Comment (amantri) [reply src/pages/paras/[user_id]/index.tsx:47]: We are "deleting" the para from the CM's list . Archive has a connotation of storing something such the archived entity can be later retrieved/un-archived. That is not the case here.

In general, the backend model shouldn't leak into the FE user experience.

src/backend/routers/paras.test.ts
amantri: Creating another entry in the db, either in this test (say using createPara) or in the seed db sounds good. In general, entities should have the lowest privilege needed for fulfilling their duties.

src/pages/paras/[user_id]/index.tsx
Comment (hieungo89) [reply src/pages/paras/[user_id]/index.tsx]: I changed the state to `unassignPrompt` and fixed all the code to refer to `unassign` instead of `archive`.

hieungo89:
This PR is no longer needed. The issue has been taken care of in #127 and #128. Both have been merged successfully.

PR #111, opened 2023-06-09
Create IEP, goal and subgoal tables, and insert IEP from the FE

amantri:


src/pages/_app.tsx
```diff
@@ -29,6 +29,13 @@ export default function App({
   const [trpcClient] = useState(() =>
     trpc.createClient({
       links: [
+        // Log in development and only log errors in production
+        loggerLink({
+          enabled: (opts) =>
+            (process.env.NODE_ENV === "development" &&
+              typeof window !== "undefined") ||
+            (opts.direction === "down" && opts.result instanceof Error),
```
codetheweb, line 37: ideally this would use the existing logger we have defined in `backed/lib/logger.ts`, I think we should be able to pass the export of that to a `console` option here?

https://trpc.io/docs/client/links/loggerLink

package.json
```diff
@@ -37,6 +37,7 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-timer-hook": "^3.0.6",
+    "superjson": "^1.12.3",
```
codetheweb, line 40: it looks like superjson is mostly useful for serializing/deserializing Dates? would be good to include a comment somewhere documenting why it's used

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -56,3 +56,23 @@ CREATE TABLE "file" (
...
+  goal_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(), -- TODO: add index to allow reordering
+  iep_id UUID REFERENCES "iep" (iep_id),
+  description TEXT
+);
+
+CREATE TABLE "subgoal" (
+  subgoal_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(), -- TODO: add index to allow reordering
+  goal_id UUID REFERENCES "goal" (goal_id),
+  description TEXT -- TODO: add more fields
+);
```
codetheweb, line 78: would recommend adding `created_at` to all tables (see above for an example) and adding `NOT NULL` constraints where appropriate

src/backend/routers/iep.test.ts
codetheweb, line 1: let's resolve the lint warnings here

(might be good to change this rule from a warning to an error)

src/backend/tests/fixtures/get-test-server.ts
```diff
@@ -60,27 +61,28 @@ export const getTestServer = async (
...
-    };
-  } else if (authenticateAs === "admin") {
+    case "case_manager":
+      user = seed.case_manager;
+      break;
+
+    case "para":
+      user = seed.para;
+      break;
+  }
```
codetheweb, line 77: 🐧 a simple lookup could work here:
```ts
const user = seed[authenticateAs]
```

src/components/Goal.tsx
```diff
@@ -0,0 +1,70 @@
+import React from "react";
+import Subgoals from "./Subgoal";
+import { trpc } from "@/client/lib/trpc";
+import Link from "next/link";
```
codetheweb, line 4:
```suggestion
```

src/components/Goal.tsx
```diff
@@ -0,0 +1,70 @@
...
+import styles from "../styles/Home.module.css";
```
codetheweb, line 5: CSS modules should be scoped to a component, would recommend adding a new module for this component

src/pages/students/[student_id].tsx
```diff
@@ -16,16 +16,36 @@ const ViewStudentPage = () => {
...
+    student_id: student_id as string,
+  });
 
+  const archiveMutation = trpc.student.unassignStudent.useMutation();
   const archiveStudent = async () => {
     if (!student) {
       return;
     }
-    mutate({ student_id: student.student_id });
+    archiveMutation.mutate({ student_id: student.student_id });
```
Comment (codetheweb) [inline src/pages/students/[student_id].tsx:28]: discussed this with @hieungo89, there's two mutation APIs in React Query:
- implicit async (used here): `.mutate()` will return `void` immediately and other parts of the return value of `.useMutation()` can be used to display success or error states. Super easy to use when your UI can display the current state.
- explicit async: `.mutateAsync()` will return `Promise<void>` and either resolve or throw like a normal Promise when the mutation completes. This is better to use for control flow than `.mutate()`, as currently in the snippet above the user will be redirected no matter what.

```suggestion
    await archiveMutation.mutateAsync({ student_id: student.student_id });
```

src/pages/iep/[iep_id].tsx
```diff
@@ -0,0 +1,76 @@
...
+  iep_id: string | string[] | undefined;
+}
+
+const Iep: React.FC<IepProps> = ({ iep_id }) => {
+  const router = useRouter();
+  const utils = trpc.useContext();
+
+  if (!iep_id) {
+    iep_id = router.query.iep_id;
+  }
```
Comment (codetheweb) [inline src/pages/iep/[iep_id].tsx:17]: function parameters generally shouldn't be modified

```suggestion
  const {iep_id} = router.query
```

src/pages/iep/[iep_id].tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useRouter } from "next/router";
+import Link from "next/link";
+import styles from "../../styles/Home.module.css";
+import Goals from "../../components/Goal";
+
+interface IepProps {
+  iep_id: string | string[] | undefined;
+}
+
+const Iep: React.FC<IepProps> = ({ iep_id }) => {
```
Comment (codetheweb) [inline src/pages/iep/[iep_id].tsx:11]: Next.js pages receive no props unless a SSR handler is defined

```suggestion
const Iep = () => {
```

src/pages/iep/[iep_id].tsx
```diff
@@ -0,0 +1,76 @@
...
+  if (!iep_id) {
+    iep_id = router.query.iep_id;
+  }
+
+  // TODO: When doing a full page reload ignoring cache (cmd+shift+r), two trpc calls are issued - one with empty
+  // "iep_id" and a second with the param populated. Figure out how to avoid the error.
+  // https://github.com/vercel/next.js/discussions/11484
+  const { data: goals, isLoading } = trpc.iep.getGoals.useQuery({
+    iep_id: iep_id as string,
+  });
```
Comment (codetheweb) [inline src/pages/iep/[iep_id].tsx:24]:
```suggestion
  const { data: goals, isLoading } = trpc.iep.getGoals.useQuery({
    iep_id: iep_id as string,
  }, {
    enabled: Boolean(iep_id)
  });
```

src/backend/db/migrations/1_initial-migrations.sql
amantri: Good suggestions. Added `created_at` and `NOT NULL` constraints. At some point, we should probably implement the full history for our tables.

package.json
amantri: Added a comment to backend/trpc.ts.

src/components/Goal.tsx
amantri: Addressed the lints

src/backend/routers/iep.test.ts
amantri: Resolved the lints. Agree on making the lint check to fail even on warnings. Folks can explicitly disable a particular lint if needed.

src/pages/iep/[iep_id].tsx
Comment (amantri) [reply src/pages/iep/[iep_id].tsx:24]: Thanks!

src/pages/iep/[iep_id].tsx
Comment (amantri) [reply src/pages/iep/[iep_id].tsx:11]: Ah ok - good to know. Done

src/pages/iep/[iep_id].tsx
Comment (amantri) [reply src/pages/iep/[iep_id].tsx:17]: Done

src/pages/students/[student_id].tsx
Comment (amantri) [reply src/pages/students/[student_id].tsx:28]: Done

src/components/Goal.tsx
amantri: Good point. Added new modules for the components. What is the best way to share styles across components?

codetheweb: generally if styles need to be shared across components that's a good indication that there should be a new parent component that encapsulates the shared styles

codetheweb (approved):

PR #112, opened 2023-06-09
Nodemailer test

BrettEastman:
- Add nodemailer-mock dependency
- Update nodemailer.ts to getTransporter arrow function
- Update paras.ts to use getTransporter with req.ctx.env
- Add createPara test in paras.test.ts file
- Update get-test-server.ts to add mockery, add nodemailerMock to return.


package.json
```diff
@@ -24,13 +24,16 @@
...
+    "@types/mockery": "^1.4.30",
     "axios": "^1.3.2",
     "dotenv": "16.0.3",
     "kysely": "^0.23.4",
+    "mockery": "^2.1.0",
     "ms": "^2.1.3",
     "next": "13.1.6",
     "next-auth": "^4.20.1",
     "nodemailer": "^6.9.2",
+    "nodemailer-mock": "^2.0.1",
```
codetheweb, line 36: let's move these to `devDependencies` since they're not needed to run the application

src/backend/routers/paras.test.ts
```diff
@@ -18,3 +20,27 @@ test("getParaById", async (t) => {
...
+  t.truthy(
+    await db
+      .selectFrom("user")
+      .where("first_name", "=", "Foo")
+      .selectAll()
+      .executeTakeFirst()
+  );
+
+  nodemailerMock.mock.getSentMail();
+  console.log(nodemailerMock.mock.getSentMail());
```
codetheweb, line 45: I would add an assertion something like
```suggestion
  t.true(nodemailerMock.mock.getSentMail().some(mail => mail.subject.includes("confirmation")))
```

src/backend/tests/fixtures/get-test-server.ts
```diff
@@ -1,3 +1,10 @@
+import mockery from "mockery";
+import * as nodemailerMock from "nodemailer-mock";
+mockery.enable({
+  warnOnUnregistered: false,
+});
+mockery.registerMock("nodemailer", nodemailerMock);
```
codetheweb, line 6: 🐧 could move this into a sibling `mocks.ts` and `import {nodemailerMock} from "./mocks"` to make it slightly cleaner

src/backend/tests/fixtures/mocks.ts
```diff
@@ -0,0 +1,8 @@
+import mockery from "mockery";
+import * as nodemailerMock from "nodemailer-mock";
+mockery.enable({
+  warnOnUnregistered: false,
+});
+mockery.registerMock("nodemailer", nodemailerMock);
+
+export default nodemailerMock;
```
codetheweb, line 8: 🐧 I'd probably do
```suggestion
export {nodemailerMock}
```
here instead since this file is called `mocks.ts` (more generic)

codetheweb (approved):
awesome!

Issue #98, opened 2023-06-05
Para flow: Create counter component [Eng]


amantri:
Create a counter component that can be instantiated multiple times for the different use cases in the data collection flow.

The component should take color/style, labels, min/max settings etc as inputs, and should expose the counter variable as an output.

![Image](https://github.com/sfbrigade/compass/assets/829820/f9ceed5e-c0d1-4f5d-9113-6dd5050a0507)

Issue #94, opened 2023-06-01
Para flow: Create timer component for data collection [Eng]


amantri:
Mocks:

Figma: https://www.figma.com/file/m09znscRXNqSziAiEbNLTf/Compass-Wireframes?type=design&node-id=4260-73489&t=duBg5z5djAaQ44DF-4

<img src="https://github.com/sfbrigade/compass/assets/829820/0c403d30-bc65-4481-95a5-6268482d05c3" width=40%>

PR #119, opened 2023-06-12
Counter and timer components

jonahchoi:
Issue #98  and #94 .

Added counter component and timer/timer input. Currently sitting in the trials pathway for display purposes.


src/components/para_trials/counter.tsx
```diff
@@ -0,0 +1,60 @@
...
+    if (count === maxCount) {
+      return;
+    }
+    setCount(count + 1);
+  };
+  const decrementCount = () => {
+    if (count === minCount) {
+      return;
+    }
+    setCount(count - 1);
```
codetheweb, line 25: 🐧 mostly personal style preference but could do 
```suggestion
    setCount(Math.max(count - 1, minCount));
```
here instead (and likewise above)

src/components/para_trials/counter.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useState } from "react";
+import styles from "../../styles/Paraflow.module.css";
+
+interface CounterProps {
+  title: string;
+  maxCount: number;
+  minCount: number;
+  color: string;
```
codetheweb, line 8: this should probably be a const union, so something like
```suggestion
  color: "yellow" | "green" | "blue";
```

to help whoever's calling it realize the correct values (types can be a great way to help document code)

src/components/para_trials/timerInput.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { useState } from "react";
+import styles from "../../styles/Paraflow.module.css";
+
+interface TimerInput {
+  onStartTimer: (inputTimeInStandard) => void;
```
codetheweb, line 5: I would recommend that this component internally stores and provides durations as milliseconds/seconds rather than a string so callers don't have to worry about conversion

src/components/para_trials/counter.tsx
jonahchoi: I wasn't sure how we would be implementing color choices, so I left it open for future changes. But this is a great tip that I'll add in for now. Thanks!

src/components/para_trials/timerInput.tsx
jonahchoi: In order to store the time as a number, I would have to convert the user input on every change of the input string, instead of once when the "start" button is pressed. I could create a function within the component to convert the string to seconds, before being passed to onStartTimer. Please let me know your thoughts on this.

codetheweb: yeah converting before passing to `onStartTimer()` seems good

codetheweb (commented):
looking good! looks like there's a few lint warnings to address

codetheweb (approved):
nice

PR #125, opened 2023-06-15
Log trpc calls

amantri:
All calls are logged in development, and only errors are logged in production.

Screenshot 2023-06-15 at 2 57 20 PM

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -3,7 +3,7 @@ import { SessionProvider } from "next-auth/react";
 import type { AppProps } from "next/app";
 import { trpc } from "client/lib/trpc";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import { httpBatchLink } from "@trpc/client";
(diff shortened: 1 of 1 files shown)


amantri:
> ideally this would use the existing logger we have defined in `backed/lib/logger.ts`, I think we should be able to pass the export of that to a `console` option here?
> 
> [trpc.io/docs/client/links/loggerLink](https://trpc.io/docs/client/links/loggerLink)

The Winston logger in backend/lib is the backend logger, while this is logging in the browser console (see screenshot above). I got a module-not-found error when I tried to import it in `_app.tsx`. Can we share a logger between the FE and the BE?

codetheweb:
> The Winston logger in backend/lib is the backend logger, while this is logging in the browser console (see screenshot above). I got a module-not-found error when I tried to import it in _app.tsx. Can we share a logger between the FE and the BE?

oh my bad, thought it was logging in the server process for some reason

neat!

codetheweb (approved):

Issue #84, opened 2023-05-26
Add global error handler for trpc


codetheweb:
There should be a global error handler passed as an option to the trpc provider in `src/pages/_app.tsx` that shows a toast or alert on API errors.

shin101:
Before writing any code I wanted to do a sanity check and make sure my understanding is aligned with what this task is looking for...

I took at look at `src/pages/_app.tsx` as well as [react-toastify](https://fkhadra.github.io/react-toastify/positioning-toast) we could potentially use.

I could return something like <trpc.Provider client={trpcClient} queryClient={queryClient} **onError={handleError}**> and create a separate handleError function that shows toast error message

Did I understand this correctly?

codetheweb:
MUI has a snackbar/toast component built in, so I would use that: https://mui.com/material-ui/react-snackbar (once https://github.com/sfbrigade/compass/pull/124 is merged)

Otherwise yes, sounds good!

shin101:
Currently if a user hits an endpoint that does not exist (localhost:3000/hi for instance) they will be greeted with an error toast. Is this close to what we had in mind? I am happy to demo during the eng sync as well. 

![Image](https://github.com/sfbrigade/compass/assets/44033225/a1b61e5e-b6bb-4379-ac8f-a65631e8e7a9)

codetheweb:
No, that might be useful but this issue was originally for displaying an error when a trpc mutation fails.

shin101:
**Quick update** : 
Configured queryCache with an onError handler which gets triggered whenever an error occurs during a query or mutation. Seems to work correctly!

<img width="1452" alt="Screenshot 2023-07-04 at 10 16 52 AM" src="https://github.com/sfbrigade/compass/assets/44033225/65f829b9-a59a-458e-8cd8-47fb04e67335">

With existing mutations like creating a new para or a student for instance, currently we are console logging the error but would it make sense for me to change existing code in MyStudents.tsx and MyParas.tsx so that the toast is displayed instead?

codetheweb:
awesome, yes I'd change/remove the current `onError` handlers

shin101:
Update : will be refactoring this code quite a bit since there are new design requirements that the current toast library cannot handle

PR #127, opened 2023-06-21
Allow assigning paras to case managers

hieungo89:
Addressing issue #72.

This is archiving Para flow version 2. We had many merge conflicts with the previous archivePara flow, so starting anew with the main branch is better. We also made significant progress with this new flow. I've shared the things completed here below.

- [x]  Added relational table for cm and para
- [x]  Added method `assignParaToCaseManager`
         This checks the db for the para using the email
         - If para is not in the db, it will call the `createPara` method
         - If para is in the db, it will add both names to the relational table
- [x]  Modified `getAllParas` to `getMyParas`

Other tasks to be completed after this:
- [ ] Unassign Para flow
- [ ] Para tests

Other contributors include @KCCPMG, @patrickwchoi, & @tessathornberry.


src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -10,8 +10,11 @@ CREATE TABLE "user" (
   image_url TEXT
 );
 
--- table for CM's para list will go here
-
+CREATE TABLE "paras_assigned_to_case_manager" (
+  case_manager_id UUID REFERENCES "user" (user_id) ON DELETE SET NULL,
```
codetheweb, line 14: these rows are only meaningful if both columns have a value, so I would recommend adding a `NOT NULL` check and dropping the row on relational deletes

src/backend/routers/para.ts
```diff
@@ -54,5 +69,41 @@ export const para = router({
       return result;
     }),
 
+  /**
+   * The function called in the frontend.
+   * Checks if para exists and finds it. If it doesn't we create one using createPara(). Then we assign the para to CM.
+   * Sends email only if createPara() is called. Not sure if this needs to be changed.
+   */
+  assignParaToCaseManager: authenticatedProcedure
```
codetheweb, line 77: how will this be used in the UI? I'm thinking we might want to reduce the scope of this to assign a para to a case manager given an email or para ID and remove the creation logic to make it a little easier to understand

amantri: Agree. Ideally the signature of this method should be:
assignParanToCaseManager(para_id, case_manager_id)

The para creation can be handled in the function that calls it. This is the single responsibility principle of functions.

src/backend/db/migrations/1_initial-migrations.sql
hieungo89: I've taken this out in the upcoming commit.

src/components/case_manager/AllParas.tsx
```diff
@@ -6,22 +6,23 @@ import PersonCreationForm from "./PersonCreationForm";
 
 const AllParasPage = () => {
```
amantri, line 7: Consider renaming this to MyParas (also, the Page suffix is not needed - it it technically a react component, and can in principle be used in other contexts)

hieungo89: I've renamed it in this upcoming push.

hieungo89:
`AllParas.tsx` is now converted into `MyParas.tsx`. However, I still kept the `AllParas.tsx` file. Should it be kept or removed?

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -10,8 +10,11 @@ CREATE TABLE "user" (
   image_url TEXT
 );
 
--- table for CM's para list will go here
-
+CREATE TABLE "paras_assigned_to_case_manager" (
+  case_manager_id UUID REFERENCES "user" (user_id),
```
codetheweb, line 14: 🐧 NOT NULL check on both columns would still be helpful

src/backend/routers/para.ts
```diff
@@ -17,8 +17,33 @@ export const para = router({
...
-  getAllParas: authenticatedProcedure.query(async (req) => {
-    const result = await req.ctx.db.selectFrom("user").selectAll().execute();
+  getParaByEmail: authenticatedProcedure
+    .input(z.object({ email: z.string() }))
+    .query(async (req) => {
+      const { email } = req.input;
+
+      const result = await req.ctx.db
+        .selectFrom("user")
+        .where("email", "=", email.toLowerCase())
```
codetheweb, line 27: 🐧 could put this on the zod input schema instead: https://github.com/colinhacks/zod#transform

src/components/case_manager/MyParas.tsx
```diff
@@ -4,24 +4,38 @@ import styles from "../../styles/Dashboard.module.css";
...
-  const { mutate } = trpc.para.createPara.useMutation({
-    onSuccess: () => {
-      return utils.para.getAllParas.invalidate();
+  const { data: paras, isLoading } = trpc.para.getMyParas.useQuery();
+
+  const createPara = trpc.para.createPara.useMutation({
+    onSuccess: () => utils.para.getMyParas.invalidate(),
+    onSettled: (data, error) => {
+      if (error) {
+        return alert(error.message);
```
codetheweb, line 15: 🐧 maybe log to console instead for now? hopefully https://github.com/sfbrigade/compass/issues/84 is in soon

src/backend/routers/para.ts
```diff
@@ -29,30 +54,55 @@ export const para = router({
...
-        .values({ first_name, last_name, email, role })
+        .values({
+          first_name,
+          last_name,
+          email: email.toLowerCase(),
+          role: "staff",
+        })
+        // TODO: make a better way to handle email conflict when adding staff that is already in db.
+        .onConflict((oc) =>
+          oc.column("email").doUpdateSet({ email: email.toLowerCase() })
```
codetheweb, line 72: the update here is a no-op since by definition the email will be the same, would recommend doing `.doNothing()` instead for clarity

hieungo89: If `.doNothing()` is used, then the result won't be able to generate data to return to frontend. What if I do an additional check to the db. and only create new para if the para is not there?

src/backend/routers/para.ts
hieungo89: I'm not sure how to do this. Maybe you can show me later, we can do this in another PR, or I can remove this completely since it's not being used anywhere yet.

codetheweb (approved):
some tests for the new procedures might be helpful, otherwise looks great!

hieungo89:
> some tests for the new procedures might be helpful, otherwise looks great!

Tests will come with the next PR when the unassignPara flow is added.

PR #128, opened 2023-06-24
Unassign Para and Para Flow Tests

hieungo89:
Addressing issue #72 Part 2 of #127.

- [x]  Unassign Para flow
- [x]  Para tests

Other contributors include @KCCPMG, @patrickwchoi, & @tessathornberry.

### DIFF
--- a/src/backend/routers/para.test.ts
+++ b/src/backend/routers/para.test.ts
@@ -1,5 +1,5 @@
 import test from "ava";
-import { getTestServer } from "backend/tests";
+import { getTestServer } from "@/backend/tests";
 
 test("getParaById", async (t) => {
   const { trpc, db } = await getTestServer(t, { authenticateAs: "para" });
--- a/src/backend/routers/para.test.ts
+++ b/src/backend/routers/para.test.ts
@@ -81,7 +81,7 @@ test("createPara", async (t) => {
   );
 });
 
-test("doNotAddDuplicateEmail", async (t) => {
+test("paras are de-duped by email", async (t) => {
   const { trpc, db } = await getTestServer(t, { authenticateAs: "para" });
 
   t.falsy(await trpc.para.getParaByEmail.query({ email: "foo.bar@email.com" }));
--- a/src/backend/routers/para.test.ts
+++ b/src/backend/routers/para.test.ts
@@ -7,9 +7,9 @@ test("getParaById", async (t) => {
   const { user_id } = await db
     .insertInto("user")
     .values({
-      first_name: "John",
-      last_name: "Doe",
-      email: "john.doe@example.com",
+      first_name: "Foo",
+      last_name: "Bar",
+      email: "foo.bar@email.com",
       role: "staff",
     })
     .returningAll()
@@ -25,9 +25,9 @@ test("getParaByEmail", async (t) => {
   const { email } = await db
     .insertInto("user")
     .values({
-      first_name: "John",
-      last_name: "Doe",
-      email: "john.doe@example.com",
+      first_name: "Foo",
+      last_name: "Bar",
+      email: "foo.bar@email.com",
       role: "staff",
     })
     .returningAll()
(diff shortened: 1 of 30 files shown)


src/backend/routers/para.test.ts
```diff
@@ -44,3 +77,52 @@ test("createPara", async (t) => {
...
+
+test("assignParaToCaseManager", async (t) => {
+  const { trpc, seed } = await getTestServer(t, { authenticateAs: "para" });
+
+  await trpc.para.assignParaToCaseManager.mutate({
+    para_id: seed.paraJohn.user_id,
+  });
+
+  const myParas = await trpc.para.getMyParas.query();
+  t.is(myParas.length, 1);
```
codetheweb, line 105: an assertion before `.assignParaToCaseManager()` would be good to confirm that this is initially `0` (often useful to make sure seed data isn't affecting tests)

src/backend/routers/para.test.ts
```diff
@@ -44,3 +77,52 @@ test("createPara", async (t) => {
...
+      .insertInto("user")
+      .values({
+        first_name: "Foos",
+        last_name: "Bar",
+        email: "john@example.com",
+        role: "staff",
+      })
+      .execute();
+  });
+});
```
codetheweb, line 95: it's not entirely clear what this is testing at a quick glance

if I was writing this test, I would:
- replace `john@example.com` with `seed.para.email` or whatever's returned from the fixture seed (I think this is why it's throwing?)
- edit the test name to make the goal more clear
- move this test to something like `src/backend/db/migrations/1_initial-migrations.test.ts`, since, as written, this isn't touching any trpc procedures. Alternatively switch out the database call for a procedure call

src/backend/tests/seed.ts
```diff
@@ -25,8 +25,21 @@ export const seed = async (db: KyselyDatabaseInstance) => {
     .returningAll()
     .executeTakeFirstOrThrow();
 
+  //! For tests, is it better to insert into database this para or have a seed of this para already in the db?
```
codetheweb, line 28: why can't we use the existing para above?

hieungo89: In the test, we're using the authentication as `para`. In order to use the existing para above, we have to authenticate as `admin` right?

src/backend/routers/para.test.ts
hieungo89: This test is to test for non-duplicate email for `user` in the db. We can move it to that route if it's more appropriate.

src/backend/tests/seed.ts
codetheweb: sorry, not understanding

passing `para` to the test fixture authenticates as the above para:

Screenshot 2023-06-24 at 9 02 02 PM

(we should probably change this to `staff` to match the role but that can be a separate PR)

src/backend/routers/para.test.ts
codetheweb: I think if we address the first two points above and swap the database call for a procedure call we should be good

(the convention in this repo should be to write as high-level tests as possible that replicate how the frontend interacts with the data persistence layer)

codetheweb (dismissed):

hieungo89:
The commit branch 17b639d was automatically added after I pulled and rebase branch with main. The error is fixed in commit branch 73c49bf.

src/backend/routers/para.test.ts
```diff
@@ -44,3 +80,65 @@ test("createPara", async (t) => {
       .some((mail) => mail.subject?.includes("confirmation"))
   );
 });
+
+test("doNotAddDuplicateEmail", async (t) => {
```
codetheweb, line 84:
```suggestion
test("paras are de-duped by email", async (t) => {
```

codetheweb (approved):

src/backend/routers/para.test.ts
```diff
@@ -44,3 +80,65 @@ test("createPara", async (t) => {
       .some((mail) => mail.subject?.includes("confirmation"))
   );
 });
+
+test("paras are de-duped by email", async (t) => {
```
amantri, line 84: /
```suggestion
test("paras are deduped by email", async (t) => {
```

amantri (approved):

PR #129, opened 2023-06-25
Use consistent path aliases

### LINKED ISSUE #56: Consistent path aliases
Either `backend/*` or `@/backend/*`, not both.

codetheweb:
Closes #56.

### DIFF
--- a/src/backend/context.ts
+++ b/src/backend/context.ts
@@ -1,5 +1,5 @@
 import { CreateNextContextOptions } from "@trpc/server/adapters/next";
-import { getDb } from "backend/db/lib/get-db";
+import { getDb } from "@/backend/db/lib/get-db";
 import { S3Client } from "@aws-sdk/client-s3";
 import { Session, getServerSession } from "next-auth";
 import { Env } from "./lib/types";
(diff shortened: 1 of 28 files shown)


src/components/case_manager/MyParas.tsx
```diff
@@ -1,4 +1,4 @@
-import { trpc } from "client/lib/trpc";
+import { trpc } from "@/client/lib/trpc";
 import React from "react";
 import styles from "../../styles/Dashboard.module.css";
```
amantri, line 3: 🐧 Change this to use @/styles/.... Here and elsewhere

src/backend/context.ts
```diff
@@ -1,5 +1,5 @@
 import { CreateNextContextOptions } from "@trpc/server/adapters/next";
-import { getDb } from "backend/db/lib/get-db";
+import { getDb } from "@/backend/db/lib/get-db";
```
amantri, line 2: OOC: will an import like "backend/db/lib/get-db" (w/o @) now give an error?

amantri (approved):

src/backend/context.ts
codetheweb: yep

Issue #108, opened 2023-06-09
CM flow: Reconfigure CM Dashboard: Staff page. [ENG]


tessathornberry:
## Behaviors of the updated Staff page:

- [ ] Contains an Add Staff component prompting requirements for their first name, last name, e-mail address, and (optional)phone number, with an "Add Staff" button
- [ ] Once any staff have been added, their first and last names, email addresses and phone numbers(if inputted) render below the "add staff" component, sorted by order entered or perhaps by last name, if you wish.
- [ ] clicking on a staff member's name links to that staff member's page in pages/paras/user_id

- [ ] Optionally render the total number of staff, as in the example, default 1.

### This issue will likely require restructuring between the pages and components files.
Most of the functions and mapping have been figured out, and are available for copying and/or review from components/case_manager/AllParas
#### Rather than renaming, moving, or deleting files, make new files. You can always comment out the original files after copying usable code. 
You want to be sure your branch works before any deletion or alteration of the original files.
Once your branch works, you can delete the original file (e.g. in src/components if you are instead using a file in src/pages) if you are ready to push your PR.

States:

![Image](https://user-images.githubusercontent.com/105292208/244580813-cd644feb-6aba-4f73-97bc-9506d401ada3.png)

![Image](https://user-images.githubusercontent.com/105292208/244580946-ff0ac068-d12f-4998-8780-e841fdca7528.png)

![Image](https://user-images.githubusercontent.com/105292208/244581053-54a79ff4-566e-4bf7-96a7-d3f7d622ba75.png)

codetheweb:
Blocked by #106.

tessathornberry:
This is no longer blocked - it has been revised to be open to assignment.

Issue #107, opened 2023-06-08
CM flow: Reconfigure Current CM dashboard: Students page. [ENG]


tessathornberry:
## Behaviors of the updated Students page:

- [x] Contains an Add Students component prompting requirements for their first name, last name, and e-mail address, with an "Add Student" button
- [x] Once any students have been added, their first and last names (and possibly e-mail addresses) render below the "add student" component, sorted by order entered or perhaps by last name, if you wish.
- [x] clicking on a student's name links to that student's page in pages/student/student_id

- [x] Optionally render the total number of students, as in the example, default 0.

### This issue will likely require restructuring between the pages and components files.
Most of the functions and mapping have been figured out, and are available for copying and/or review from components/case_manager/AllStudents
#### Rather than renaming, moving, or deleting files, please make new files. You can always comment out the original files after copying usable code. 
You want to be sure your branch works before any deletion or alteration of the original files.
Once your branch works, you can delete the original file (e.g. in src/components if you are instead using a file in src/pages) if you are ready to push your PR.

States:

![Image](https://github.com/sfbrigade/compass/assets/105292208/90096509-95e4-4de8-90c8-9414e67baacc)

(imagine the students and e-mails are unique!)
![Image](https://github.com/sfbrigade/compass/assets/105292208/4075e471-983c-4476-b144-ccca6148af62)

(please ignore table contents and focus on button placement when the "Add Students" component is not visible:
![Image](https://github.com/sfbrigade/compass/assets/105292208/5c2674f2-04b5-4daa-8547-9ae9ca8f6e88)

codetheweb:
Blocked by #106.

tessathornberry:
This is no longer blocked - it has been revised to be open to assignment.

PR #137, opened 2023-06-29
Student/Staff pages #107/#108

jonahchoi:
#107 #108 : Added table component to render students and staff.


package.json
```diff
@@ -17,8 +17,13 @@
     "lint:fix": "eslint . --fix"
   },
   "dependencies": {
-    "@aws-sdk/client-s3": "^3.360.0",
-    "@aws-sdk/s3-request-presigner": "^3.360.0",
+    "@aws-sdk/client-s3": "^3.341.0",
```
amantri, line 20:
```suggestion
    "@aws-sdk/client-s3": "^3.360.0",
```

package.json
```diff
@@ -17,8 +17,13 @@
...
+    "@aws-sdk/s3-request-presigner": "^3.341.0",
```
amantri, line 21:
```suggestion
    "@aws-sdk/s3-request-presigner": "^3.360.0",
```

package.json
amantri: Note: you will have to update package-lock.json in this PR after this change.

src/components/case_manager/MyParas.tsx
```diff
@@ -41,24 +40,68 @@ const MyParas = () => {
...
+      label: "Last Name",
+      hasInput: true,
+    },
+    {
+      id: "email",
+      label: "Email",
+      hasInput: true,
+    },
+    {
+      id: "phone",
```
amantri, line 60: Let's only keep the first name, last name and email columns for now. We can add more columns to this table over time as we start collecting more data.

src/components/case_manager/MyParas.tsx
```diff
@@ -41,24 +40,68 @@ const MyParas = () => {
...
+  return (
+    <>
+      <PersonTable
+        people={paras as Para[]}
+        onSubmit={handleSubmit}
+        headCells={headCells}
+        type="Staff"
+      />
+    </>
+    // <div>
```
amantri, line 90: Please delete these lines

src/components/case_manager/MyStudents.tsx
```diff
@@ -28,25 +27,40 @@ const MyStudentsPage = () => {
...
+      label: "Last Name",
+      hasInput: true,
+    },
+    {
+      id: "email",
+      label: "Email",
+      hasInput: true,
+    },
+    {
+      id: "dateAdded",
```
amantri, line 47: Ditto - remove this column for now.

src/components/case_manager/types/table.ts
```diff
@@ -0,0 +1,45 @@
+export interface Person {
+  first_name: string;
+  last_name: string;
+  email: string;
+  phone?: string;
```
amantri, line 5: Remove phone and dateAdded for now. In general, it is easier to add fields/data to interfaces later than to remove, so err on the side of minimalism.

Also remove other non-essential fields in the other interfaces

src/components/case_manager/types/table.ts
```diff
@@ -0,0 +1,45 @@
+export interface Person {
```
amantri, line 1: It would be useful to have these interfaces in a more central location. 
@codetheweb: any ideas on where to put them?

src/components/para_trials/timer.tsx
amantri, line 1: Move this and related files to a different PR?

src/components/case_manager/types/table.ts
jonahchoi: Fixed everything except this one. We can either wait for a response from Max or change it later. What do you think?

amantri (approved):

src/components/case_manager/MyParas.tsx
```diff
@@ -41,24 +40,33 @@ const MyParas = () => {
     return <div>Loading...</div>;
   }
 
-  return (
-    <div>
-      <PersonCreationForm title={"Add a Para"} onSubmit={handleSubmit} />
+  const headCells: HeadCell[] = [
```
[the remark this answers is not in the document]
jonahchoi: No, its just what MUI had defined in their example, which I copied over from the docs. Since the entries are called TableCell, they used HeadCell for the header. I'm fine with renaming it if you prefer.

src/components/case_manager/PersonTable.tsx
```diff
@@ -0,0 +1,445 @@
...
+  order: Order,
+  orderBy: UserKeys
+): <T extends Student | Para>(a: T, b: T) => number {
+  return order === "desc"
+    ? (a, b) => descendingComparator(a, b, orderBy)
+    : (a, b) => -descendingComparator(a, b, orderBy);
+}
+
+// Since 2020 all major browsers ensure sort stability with Array.prototype.sort().
+// stableSort() brings sort stability to non-modern browsers (notably IE11). If you
```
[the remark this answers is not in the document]
jonahchoi: Sounds good! I'll remove stableSort and replace it with Array.sort.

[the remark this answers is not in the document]
codetheweb: The bundling built into Next.js already adds some polyfills, and we could support IE if we really wanted by changing a config parameter: https://nextjs.org/docs/architecture/supported-browsers

In general we can use the most recent ES features and not worry about compatibility.

src/components/case_manager/types/table.ts
codetheweb: all the types in this file are already present in `src/backend/db/zapatos/schema.d.ts`

we can remap them slightly if necessary or re-export them to make things more ergonomic, but we need to stick with a single source of truth for database types

for example:
```ts
import { SelectableForTable } from "zapatos/schema"
type User = SelectableForTable<"user">;
```

codetheweb (changes requested):

src/components/case_manager/types/table.ts
jonahchoi: Thanks for pointing that out. Didn't realize those existed already. I'll likely import them into this file, so that the HeadCell interface can still access them, and re-export for the Table.

src/components/case_manager/types/table.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface StudentHeadCell extends HeadCell {
+  id: keyof Student;
+}
+
+export interface ParaHeadCell extends HeadCell {
+  id: keyof Para;
+}
+
+export function isStudent(person: Student | Para): person is Student {
+  return (person as Student).student_id !== undefined;
```
codetheweb, line 21: nice

src/components/case_manager/PersonTable.tsx
```diff
@@ -0,0 +1,398 @@
...
+            {showInput && (
+              <EnhancedTableInput
+                inputCells={headCells}
+                type={type}
+                onCloseInput={handleCloseInput}
+              />
+            )}
+            {visibleRows.map((row, index) => {
+              const isItemSelected = isSelected(row.email);
+              const labelId = `enhanced-table-checkbox-${index}`;
```
codetheweb, line 355: 🐧 is the prefix here used for anything?
```suggestion
              const labelId = row.email;
```

src/components/case_manager/PersonTable.tsx
```diff
@@ -0,0 +1,398 @@
...
+      <EnhancedTableToolbar
+        numSelected={selected.length}
+        totalRows={people.length}
+        type={type}
+        onOpenInput={handleOpenInput}
+      />
+      <TableContainer>
+        <Table
+          sx={{ minWidth: 750 }}
+          aria-labelledby={`Table of ${"students/paras"}`}
```
codetheweb, line 334:
```suggestion
          aria-labelledby={`Table of students/paras`}
```

src/components/case_manager/PersonTable.tsx
```diff
@@ -0,0 +1,398 @@
...
+  "&.Mui-selected": {
+    backgroundColor: "lightblue",
+  },
+}));
+
+interface EnhancedTableProps<Person, Column> {
+  people: Person[];
+  onSubmit: (event: React.FormEvent<HTMLFormElement>) => void;
+  headCells: Column[];
+  type: "Student" | "Staff";
```
codetheweb, line 250: can this be inferred based on `people`?

codetheweb (approved):

src/components/case_manager/PersonTable.tsx
jonahchoi: We could use the `isStudent` typeguard function on the first element of the `people` array to define `type`. But the issue is if there are no elements in `people` yet, how would we know to render the students or staff version?

PR #139, opened 2023-07-05
Add uploadImage button

MarkOfosu:
I have added a component that allows a user to take a picture and upload to AWS


src/pages/cmDashboard.tsx
```diff
@@ -41,6 +42,10 @@ function cmDashboard() {
           </div>
         </div>
       </main>
+      <UploadImageComponent
```
jonahchoi, line 45: Please move this component to the `trials/[trials_id]/index.tsx` page.
Also make sure you've pulled from main, since it looks like this is the old cmDashboard.

src/components/uploadImage/uploadImage.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const videoRef = useRef<HTMLVideoElement | null>(null);
+  const canvasRef = useRef<HTMLCanvasElement | null>(null);
+
+  const openUserMedia = async () => {
+    const stream = await navigator.mediaDevices.getUserMedia({ video: true });
+    const video = videoRef.current;
+    if (video !== null) {
+      video.srcObject = stream;
+      await video.play();
+    }
```
jonahchoi, line 36: After opening and then closing the camera component, it seems like the browser is still accessing the webcam (chrome shows the red recording circle on the tab). If possible, we should stop accessing the webcam after closing the camera. As a user, it would make me anxious to see a tab "recording" when there's nothing visually doing so on the screen.

src/components/uploadImage/uploadImage.module.css
```diff
@@ -0,0 +1,65 @@
...
+  max-height: 100%;
+  width: 300px; /* Adjust the width as needed */
+  height: auto;
+  position: relative;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -10%);
+}
+
+.button {
```
jonahchoi, line 35: Is this an extra copy? If so, please remove it.

jonahchoi (changes requested):

src/components/uploadImage/uploadImage.tsx
MarkOfosu: Hi Jonah, thanks for the feedback. I have added a stop camera function and it currently closes the camera anytime it is not in use anymore.

codetheweb (commented):
what are the benefits to doing this with a camera capture over the more common `<input type="file"/>` approach?

MarkOfosu:
The camera capture approach allows the user to directly and instantly take
a picture of a student's work to upload but the input file type is a file
picker which means there must be an existing image file from their device
before they can select it. From the ux designs, it looks like they are
asking for the camera capture approach.

codetheweb:
I believe you can prompt users to capture an image using native UI on mobile devices with the approach described here: https://stackoverflow.com/a/50440368

package.json
```diff
@@ -41,6 +41,7 @@
     "postgres-migrations": "^5.3.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-icons": "^4.10.1",
```
jonahchoi, line 44: MUI includes an icon package, which we already have imported. Check it out and see if they have compatible icons for you to use instead.

jonahchoi:
You also have a few lint warnings that I would recommend fixing.

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
...
+  return (
+    <div>
+      {showStart && (
+        <button
+          className={
+            startButtonDisabled
+              ? styles.startButtonDisabled
+              : styles.startButton
+          }
+          id="start-button"
```
codetheweb, line 118:
```suggestion
```

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
...
+          <input
+            type="file"
+            accept="image/*;capture=camera"
+            capture="environment"
+            onChange={handleFile}
+            className={styles.startButton}
+          />
+          <button
+            className={styles.startButton}
+            id="cancel-button"
```
codetheweb, line 136:
```suggestion
```

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
...
+      )}
+      {showUploadPicture && (
+        <>
+          <img src={capturedImage} alt="Captured" className={styles.video} />
+          <button className={styles.startButton} onClick={onClickUploadPicture}>
+            Upload
+          </button>
+          <button
+            className={styles.startButton}
+            id="cancel-button"
```
codetheweb, line 151:
```suggestion
```

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
+import React, { useState, useEffect, useCallback } from "react";
+import styles from "./uploadImage.module.css";
+import { trpc } from "@/client/lib/trpc";
+import axios from "axios";
+import { FaCamera, FaCheckCircle } from "react-icons/fa";
+
+interface UploadImageComponentProps {
+  title: string;
+  fileName: string;
+}
```
codetheweb, line 10: needs an `onUpload: (fileId: string) => void` handler as well so the parent component can receive uploaded files

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
...
+    setShowTakePicture(true);
+  };
+
+  const handleFile = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const file = event.target.files?.[0];
+    if (file) {
+      const reader = new FileReader();
+      reader.onload = () => {
+        if (typeof reader.result === "string") {
+          setCapturedImage(reader.result);
```
codetheweb, line 37: I would recommend keeping the file as a Blob here and transforming later as necessary

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
...
+    setUploading(true);
+  };
+
+  const getPresignedUrlForUpload =
+    trpc.file.getPresignedUrlForFileUpload.useMutation();
+  const finishFileUpload = trpc.file.finishFileUpload.useMutation();
+
+  const uploadImageToCloud = useCallback(async () => {
+    const { key, url } = await getPresignedUrlForUpload.mutateAsync({
+      type: "image/png",
```
codetheweb, line 70: not guaranteed to be a png, should pull from the Blob type

package.json
MarkOfosu: got it!

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,172 @@
...
+}) => {
+  const [showStart, setShowStart] = useState<boolean>(true);
+  const [showTakePicture, setShowTakePicture] = useState<boolean>(false);
+  const [showUploadPicture, setShowUploadPicture] = useState<boolean>(false);
+  const [startButtonDisabled, setStartButtonDisabled] =
+    useState<boolean>(false);
+  const [uploading, setUploading] = useState<boolean>(false);
+  const [uploadSuccess, setUploadSuccess] = useState<boolean>(false);
+
+  const [capturedImage, setCapturedImage] = useState<string>("");
```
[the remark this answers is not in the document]
MarkOfosu: ok, I will clean it up.

src/components/uploadPicture/uploadImage.tsx
MarkOfosu: I added the onUpload as a prop as asked. I also called it later in the **uploadImageToCloud**  passing  the **Key**  from the finishFileUpload mutation after a successful upload. There has to be an onUpload handler when it is called in the parent component. Is this what is expected?

package-lock.json
codetheweb, line 1: the large diff here with no accompanying `package.json` changes is strange, would recommend `git checkout main package-lock.json` to reset it

src/components/uploadPicture/uploadImage.tsx
codetheweb: almost, `onUpload()` should be called with the `file_id` returned by `finishFileUpload.mutateAsync()` instead of the intermediate `key`

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,190 @@
...
+              id="cancel-button"
+              onClick={onCancelCaptureImage}
+            >
+              Cancel
+            </button>
+          </div>
+        </div>
+      )}
+
+      {uploading && <p id="uploading">Uploading.......</p>}
```
codetheweb, line 177: 🐧 MUI comes with [loading buttons](https://mui.com/material-ui/react-button/#loading-button) that we could use here

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,190 @@
...
+const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
+  title,
+  fileName,
+  onUpload,
+}) => {
+  const [showStart, setShowStart] = useState(true);
+  const [showTakePicture, setShowTakePicture] = useState(false);
+  const [showUploadPicture, setShowUploadPicture] = useState(false);
+  const [startButtonDisabled, setStartButtonDisabled] = useState(false);
+  const [uploading, setUploading] = useState(false);
```
codetheweb, line 23: I don't think this variable is necessary
it slightly obfuscates the logic flow since control now goes through a `useEffect()`

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
+  title,
+  fileName,
+  onUpload,
+}) => {
+  const [showStart, setShowStart] = useState(true);
+  const [showTakePicture, setShowTakePicture] = useState(false);
+  const [showUploadPicture, setShowUploadPicture] = useState(false);
+  const [startButtonDisabled, setStartButtonDisabled] = useState(false);
```
codetheweb, line 22: I think this can be derived from `Boolean(capturedImage)`?

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,190 @@
...
+            accept="image/*;capture=camera"
+            capture="environment"
+            onChange={handleFile}
+            className={styles.startButton}
+          />
+          <button className={styles.startButton} onClick={onCancelTakePicture}>
+            Cancel
+          </button>
+        </>
+      )}
```
codetheweb, line 147: it looks like this requires two clicks to open the camera, could we get rid of the first step?

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,199 @@
...
+  UploadPicture,
+}
+
+const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
+  title,
+  fileName,
+  onUpload,
+}) => {
+  const [currentStep, setCurrentStep] = useState(UploadStep.Start);
+  const [uploading, setUploading] = useState(false);
```
codetheweb, line 28: this can be derived, something like
```ts
const isUploading = getPresignedUrlForUpload.isLoading || finishFileUpload.isLoading
```

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const [uploadSuccess, setUploadSuccess] = useState(false);
+  const [capturedImage, setCapturedImage] = useState<Blob | null>(null);
+
+  const fileInputRef = useRef<HTMLInputElement | null>(null);
```
codetheweb, line 32: 🐧 I think typing this as
```suggestion
  const fileInputRef = useRef<HTMLInputElement>(null);
```
works too (slightly cleaner)

src/components/uploadPicture/uploadImage.tsx
codetheweb: it looks like this still requires two clicks?

codetheweb (commented):
looking great!

src/components/uploadPicture/uploadImage.tsx
MarkOfosu: I am not sure how to get rid of the currentStep states with isUploading state which has only two values, true or false.

codetheweb: sorry, I meant `uploading` can be derived and `currentStep` can be kept as-is

src/components/uploadPicture/uploadImage.tsx
MarkOfosu: At my end, it does not require two clicks so I am not sure what might be the issue.

codetheweb: sorry, maybe this recording makes it more clear:


https://github.com/sfbrigade/compass/assets/7410405/d92425c2-4821-483c-9e73-951c183ef584

it looks like it requires two clicks until the file browser pops up?

that being said it's not the biggest deal and we can fix it later, if you [fix the lock file](https://github.com/sfbrigade/compass/pull/139#discussion_r1267503408) I think this would be good to merge

MarkOfosu: @codetheweb Oh i see. Got it. That is because I hid the file input button until the user has clicked the button. I have tried some other ways of triggering click on the file input dialog but it did not work. Maybe, ticket this additional functionality for later.

MarkOfosu: @jonahchoi could you help with this? I think this PR has taken too long lol

codetheweb: I'll go ahead and fix the lock file, merge this in, then create some new issues 👍 

thanks for the work!

codetheweb (approved):

Issue #109, opened 2023-06-09
CM flow: Make reusable search component for searching both students and staff. [ENG]


tessathornberry:
It would be ideal if issues #106, #107, and #108 were completed and merged before this one is started.

Feel free to look at src/components/case_manager/PersonCreationForm.tsx for an example of a reusable form.

- [ ] Make a form that takes in the first names, last names, and email of a list of people, and filters by a user-inputted string. Make the results clickable to that person's page in pages.
- [ ] Insert into Staff page and Student page

codetheweb:
Blocked by https://github.com/sfbrigade/compass/issues/106, https://github.com/sfbrigade/compass/issues/107, and https://github.com/sfbrigade/compass/issues/108

tessathornberry:
Unblocked and ready for development.

PR #148, opened 2023-07-09
Search Function #109

jonahchoi:
#109  Added search to table, removed row selection and made whole row clickable as requested by design team.

### DIFF
--- a/src/components/case_manager/PersonTable.tsx
+++ b/src/components/case_manager/PersonTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useCallback, useState } from "react";
 import Table from "@mui/material/Table";
 import TableContainer from "@mui/material/TableContainer";
 import TableHead from "@mui/material/TableHead";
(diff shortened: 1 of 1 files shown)


src/components/case_manager/PersonTable.tsx
```diff
@@ -348,39 +314,25 @@ export default function EnhancedTable<
...
                   hover
-                  role="checkbox"
-                  aria-checked={isItemSelected}
+                  role="link"
                   tabIndex={-1}
                   key={row.email}
-                  selected={isItemSelected}
                   sx={{ cursor: "pointer" }}
+                  onClick={() =>
+                    handleLinkToPage(
```
codetheweb, line 327: can this not be a `<Link/>`?

src/components/case_manager/PersonTable.tsx
```diff
@@ -110,28 +88,26 @@ function EnhancedTableHead<Column extends HeadCell>({
 }
 
 interface EnhancedTableToolbarProps {
-  numSelected: number;
   totalRows: number;
   type: "Student" | "Staff";
   onOpenInput: () => void;
+  searchParam: string;
```
codetheweb, line 94: a little confused what this is, if I'm understanding correctly maybe 
```suggestion
  defaultSearchValue: string;
```
is more appropriate?

jonahchoi: I wouldn't say its a default. Maybe just searchValue or inputSearchValue?

codetheweb: it looks like this component controls the state of the value and it just is provided as the initial state, which is why I suggested naming it `default*`

but could be misunderstanding

src/components/case_manager/PersonTable.tsx
jonahchoi: The <a> tag can't be a child of the table element, preventing me from wrapping the row in a Link. There are some other workarounds like changing the root component of the table to divs. 

Keep it as is: Unable to right click/open in new tab. Keep the semantics of tables.
Change it to divs: Possible issue with semantics?

src/components/case_manager/PersonTable.tsx
jonahchoi: The state value is defined on line 237 as an empty string. It is passed down to this component, where it can be changed in the TextField from 135-148. The user inputs the search parameter which will then filter the list.

src/components/case_manager/PersonTable.tsx
jonahchoi: https://stackoverflow.com/a/71370700

codetheweb: ah, I see, my bad

src/components/case_manager/PersonTable.tsx
codetheweb: 🤦 

probably oos for this PR, but maybe splitting up the components into separate files would be good

codetheweb (approved):

src/components/case_manager/PersonTable.tsx
jonahchoi: I had that thought as well. But none of the other components get reused outside of the table, which is why I liked leaving them in the same file.

PR #153, opened 2023-07-20
Split out S3 backend for Terraform

codetheweb:
A few changes I needed to make to get started.

I split out the S3 backend from the main deployment as it seems like the recommendation is to deploy it once manually rather than continously to avoid the chicken and egg problem.

Also, the Terraform backend config is now in a separate file to make it easier to use a local dev environment.

### DIFF
--- a/terraform/compass_module/main.tf
+++ b/terraform/compass_module/main.tf
@@ -1,8 +1,4 @@
 # Creates and manages the AWS resources needed for Compass.
-#
-# Also creates the following resources for the Terraform backend configuration:
-#  s3 bucket for state backup: "${var.app_name}-${var.app_environment}-tf-state"
-#  DynamoDB for locking      : "${var.app_name}-${var.app_environment}-tf-state-locking"
 
 terraform {
   required_providers {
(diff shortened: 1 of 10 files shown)


terraform/compass_module/main.tf
```diff
@@ -15,46 +11,4 @@ terraform {
 
 provider "aws" {
   region  = var.aws_region
-  profile = var.aws_user
```
codetheweb, line 18: needed to remove this to be able to run, looks like there's other ways to set this (e.x. through an environment variable)

terraform/s3-backend/.terraform.lock.hcl
codetheweb, line 1: terraform recommends that lock files be checked in

terraform/staging/main.tf
```diff
@@ -1,10 +1,6 @@
 terraform {
   backend "s3" {
-    bucket         = "compass-staging-tf-state" # Substitute environment as needed
-    key            = "terraform.tfstate"
-    region         = "us-west-2"
-    dynamodb_table = "compass-staging-tf-state-locking" # Substitute environment as needed
-    encrypt        = true
+    key = "staging/terraform.tfstate"
```
codetheweb, line 3: recommended to use the same state bucket for multiple environments

amantri (approved):

codetheweb:
Now using GCS, so closing.

Issue #90, opened 2023-06-01
Flesh out MVP data model for Compass [Eng]


amantri:
Flesh out the data model parts that are needed for the MVP CM and para flows.

Parts to flesh out:
 - structure of IEP, goals and benchmarks
 - Data collection for a trial
 - data collection scheduling (which populates tasks on the para dashboard)

Code4Dopamine:
Mermaid.js Code Test:
How to use:
- Use an Editor like [Online Mermaid (UML) Creator] (https://mermaid.live/) or install plugins for [VSCode](https://mermaid.js.org/ecosystem/integrations.html#editor-plugins), etc.
- Commit the mermaid.js file normally

> For more, visit [here](https://mermaid.js.org/) and [here(Github Blog)](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/)

Demo Below:
```mermaid
classDiagram
    Main <|-- CaseManager
    Main <|-- Para
    Main <|-- Student
    class CaseManager{
      +String uuid
      +Method1()
      +Method2()
    }
    class Para{
      +String uuid
      +Array[] Tasks
      +Method1()
    }
    class Student{
      +String uuid
      +Method1()
    }
```

PR #157, opened 2023-07-22
MVP data models for goals, subgoals(benchmarks), tasks.

jonahchoi:
#90 :
- [x] Added tables for goals, subgoals, tasks, and collections
- [x] Added routes for IEP and para
- [x] Added some tests (not fully encompassing)
- [x] Modified frontend components to display new data

Worked with: @amantri @tessathornberry @hieungo89


src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -70,12 +71,44 @@ CREATE TABLE "goal" (
   goal_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(), -- TODO: add index to allow reordering
   iep_id UUID REFERENCES "iep" (iep_id),
   description TEXT NOT NULL,
-  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  category TEXT NOT NULL CHECK (category IN ('writing', 'reading', 'math', 'other'))
```
codetheweb, line 75: 🐧 convention has been to have "common" columns like `created_at` at the end of the table definition

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -70,12 +71,44 @@ CREATE TABLE "goal" (
...
 );
 
 CREATE TABLE "subgoal" (
   subgoal_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(), -- TODO: add index to allow reordering
   goal_id UUID REFERENCES "goal" (goal_id),
   description TEXT NOT NULL,
-  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  instructions TEXT,
```
codetheweb, line 83: 🐧 I'd probably make this not null and default to `""`

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -70,12 +71,44 @@ CREATE TABLE "goal" (
...
+  target_max_attempts INTEGER --How many "questions" to administer in a single sitting
```
codetheweb, line 84: is a not null value valid here?

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -70,12 +71,44 @@ CREATE TABLE "goal" (
...
+  created_by_user_id UUID REFERENCES "user" (user_id),
+  -- TODO: Possibly add optional reference to "task"
+  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+
+  success_with_prompt INTEGER NOT NULL,
+  success_without_prompt INTEGER NOT NULL,
+
+  -- Optional depending on type of task
+  notes TEXT,
+  image_list TEXT[]
```
codetheweb, line 109: it doesn't look like this is being used yet so I might remove it
this should point to `file_id`s, it's usually a bit of an antipattern to store arrays in sql

src/components/case_manager/PersonTable.tsx
```diff
@@ -289,7 +289,14 @@ export default function EnhancedTable<
   return (
     <Box sx={{ width: "75%" }}>
       {/* Form can't be integrated with table (can't span multiple cells), so the form is on the outside with inputs referencing its id */}
-      <form onSubmit={onSubmit} id="table_input_form"></form>
+      <form
+        onSubmit={(event: React.FormEvent<HTMLFormElement>) => {
+          onSubmit(event);
```
codetheweb, line 294: type for `onSubmit` should be updated above to return `Promise<void> | void` and then we can await it here to prevent resetting the form if `onSubmit` is async and throws

src/components/case_manager/PersonTable.tsx
```diff
@@ -289,7 +289,14 @@ export default function EnhancedTable<
...
+          const form = document.getElementById("table_input_form");
+          (form as HTMLFormElement).reset();
```
codetheweb, line 296:
```suggestion
          event.target.reset()
```

codetheweb (commented):
can't speak much to the data model but implementation is looking pretty good :)

src/backend/db/migrations/1_initial-migrations.sql
jonahchoi: I agree. I was thinking of having files reference trials, but kind of left it here as a placeholder since it wasn't essential yet. I'll remove it for now.

src/backend/db/migrations/1_initial-migrations.sql
jonahchoi: I think its optional to define a target max, and adding something like `not null default 0` might be confusing/unintuitive? I guess if it was default 0, we just don't render it on the front end. What do you think?

src/components/case_manager/PersonTable.tsx
jonahchoi: Oops, I was playing around with this for Tessa's issue and forgot to remove it. But thank you for this optimization! I'll let Tessa know about it.

src/backend/db/migrations/1_initial-migrations.sql
codetheweb: if it's optional seems good to me 👍

amantri (approved):

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -45,7 +45,8 @@ CREATE TABLE "student" (
   first_name TEXT NOT NULL,
   last_name TEXT NOT NULL,
   email TEXT NOT NULL UNIQUE,
-  assigned_case_manager_id UUID REFERENCES "user" (user_id) ON DELETE SET NULL
+  assigned_case_manager_id UUID REFERENCES "user" (user_id) ON DELETE SET NULL,
+  grade SMALLINT
```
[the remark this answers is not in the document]
jonahchoi: The newest mockups have grade on the student table and on students' pages. Although I forgot to implement it elsewhere. I can either a) remove it for now and let another issue handle it or b) write a todo comment.

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -70,12 +71,40 @@ CREATE TABLE "goal" (
...
   description TEXT NOT NULL,
+  category TEXT NOT NULL CHECK (category IN ('writing', 'reading', 'math', 'other')),
   created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
 CREATE TABLE "subgoal" (
   subgoal_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(), -- TODO: add index to allow reordering
   goal_id UUID REFERENCES "goal" (goal_id),
   description TEXT NOT NULL,
+  instructions TEXT NOT NULL DEFAULT '',
```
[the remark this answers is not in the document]
jonahchoi: I actually tried this first, and found out that in SQL, single and double quotes actually have different meanings. Single quotes are for string literals, while double quotes are "delimited identifiers", meaning they reference tables or columns.

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -70,12 +71,40 @@ CREATE TABLE "goal" (
...
+  target_max_attempts INTEGER, --How many "questions" to administer in a single sitting
```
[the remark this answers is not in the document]
jonahchoi: That's one reason I left it as an optional column. But I believe we discussed with Niko that our MVP would focus on one collection type, with plans to add more in the future. Maybe we can brainstorm better options for this?

src/backend/db/migrations/1_initial-migrations.sql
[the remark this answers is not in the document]
amantri: In general, try to keep a PR self contained that is focused on a specific feature/bug. This field would ideally be in the PR that adds that to the frontend/backend, but since you have already added it here, I will leave it to your discretion if you want to leave it in or remove it :)

src/backend/db/migrations/1_initial-migrations.sql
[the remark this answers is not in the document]
amantri: Ah - thanks for the info. TIL

src/backend/db/migrations/1_initial-migrations.sql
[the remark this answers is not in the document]
amantri: SG. Let's get this PR merged in, and we can update it later.

src/backend/db/migrations/1_initial-migrations.sql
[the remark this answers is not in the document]
jonahchoi: Got it. Sometimes I think of something and just toss it in, but I'll try to keep things more self contained from now on.

src/backend/routers/iep.test.ts
```diff
@@ -32,4 +42,13 @@ test("basic flow - add/get goals and subgoals", async (t) => {
     goal_id: goal1!.goal_id,
   });
   t.is(gotSubgoals.length, 2);
+
+  t.truthy(
```
amantri, line 46: 🐧 Ideally, we should be using our API methods instead of directly querying the db. We will likely need a `getTasks` method down the line when we want to do the Para schedule view. You can add a TODO for that now.

amantri (approved):
LGTM

PR #169, opened 2023-08-03
CSS Restructuring

jonahchoi:
#150 Restructured CSS architecture using global modules and TS declaration files to enable type-safe usage. 

1. `globals.css` is only for common variables and styling that applies to ALL elements 
2. Any reusable component styling (like buttons, inputs, etc) will be in a global module.
3. Component modules will be located within the component folder.
4. Styles will be imported as `$module_name`. Ex: `import $button from "@/styles/Button.module.css"`
5. Use the `type-css` npm command to automatically create declaration files for css, and the `type-css:check` command to verify these files are correct.

Co-authored-by: Hannah McGowen @hannahmcg 
Co-authored-by: Patrick Choi @patrickwchoi


codetheweb (commented):
the `.module.css.d.ts` files add a lot of noise, what do you think about instead adding them to `.gitignore` (not checking them in) and adding a npm postinstall hook to generate them?

jonahchoi:
@codetheweb I really wanted to shove them all into their own separate folder, but couldn't find a way for TS to recognize them without being in the same folder as the base css file. 

One downside I could think of is not being able to run type-css:check automatically on PRs, but I think its possible for the github action runner to create the files and then verify types are used correctly. I'll look into it.

.github/workflows/pr.yml
```diff
@@ -12,10 +12,10 @@ jobs:
         with:
           node-version: 18
           cache: "npm"
-      - name: Install dependencies
+      - name: Install dependencies & generate css types
         run: npm i --ci
       - name: Check types
-        run: npm run type-check
+        run: npm run type:check
```
codetheweb, line 18: I think this is moot if they're not being checked in and generated within the same workflow?

codetheweb (commented):
nice

.github/workflows/pr.yml
jonahchoi: The css types are generated within the workflow, then the `type:check` (which is the regular typescript check) command will check for type errors throughout the repo. This way it'll effectively see if an imported css module is using an invalid property. 

The `type-css:check` command will still be available for us to use locally, in order to double check that our declaration files are updated.

codetheweb: oh my bad, read too quickly

codetheweb (approved):

PR #171, opened 2023-08-07
Create secrets and add instructions for Google OAuth credentials

codetheweb:
Unfortunately OAuth applications can't be created through terraform.


terraform/gcs/cloud_run.tf
```diff
@@ -32,19 +32,29 @@ resource "google_cloud_run_v2_service" "run_service" {
...
+          }
         }
       }
 
-      liveness_probe {
-        # Wait for migrations to run
-        initial_delay_seconds = 30
-        failure_threshold = 5
+      env {
+        name = "GOOGLE_CLIENT_SECRET"
```
amantri, line 46: very cool!

amantri (approved):

PR #172, opened 2023-08-07
Reconfigure pages-protected so only login page is rendered unless the user is logged in

KotovSyndrome:
- Replaced tRPC-based auth with session auth for page wrapper
- Protected all pages


src/pages/settings.tsx
```diff
@@ -8,4 +9,6 @@ const Settings = () => {
   );
 };
 
-export default Settings;
+const protectedSettings = requiresAdminAuth(Settings);
```
jonahchoi, line 12: Instead of doing this for every page, I think it would be better to put it on one component that's always available. Maybe the new Layout component?

src/client/lib/protected-page.tsx
```diff
@@ -1,25 +1,25 @@
-import { trpc } from "./trpc";
 import { useEffect } from "react";
 import { useRouter } from "next/router";
+import { useSession } from "next-auth/react";
 
 export const requiresAdminAuth =
```
jonahchoi, line 5: I believe this func was originally for admin pages. Did we discuss removing it at the last meeting? If not we should probably keep it and have a separate regular login auth.

src/pages/settings.tsx
KotovSyndrome: Yeah, that would be more aligned with DRY principles. I'll get on it!

src/client/lib/protected-page.tsx
KotovSyndrome: Gotcha, my original plan was to co-opt this one, but I'll make a separate one instead.

src/pages/settings.tsx
jonahchoi, line 12: This page still has the admin auth, but I'm not sure if settings will be for admins only? I thought settings were for any user.

src/pages/signInPage.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useRouter } from "next/router";
+import { useSession } from "next-auth/react";
+
+const SignInPage = () => {
+  const router = useRouter();
+  const { status } = useSession();
+
+  useEffect(() => {
+    status === "authenticated" ? void router.push("/students") : null;
+  });
```
jonahchoi, line 15: Ditto for dependency array.

src/pages/students/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import MyStudents from "@/components/case_manager/MyStudents";
+import MyStudents from "../../components/case_manager/MyStudents";
```
jonahchoi, line 2: Any specific reason for this change?

jonahchoi (changes requested):

src/pages/students/index.tsx
KotovSyndrome: Nope, I think this was an accidental change. Will revert.

src/client/lib/authenticated-page.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+    // Include any public pathnames here
+    const publicPaths = ["/signInPage"];
+    const pathIsProtected = !publicPaths.includes(router.pathname);
+
+    useEffect(() => {
+      if (status === "unauthenticated" && pathIsProtected) {
+        void router.push("/");
+      }
+    });
```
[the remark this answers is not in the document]
tessathornberry: Let's try it! TS prompted me to add all three dependencies in the useEffect, so it will trigger on changes in router, status, and pathIsProtected after my changes, but that's still better than every render.

src/components/Layout.tsx
```diff
@@ -1,18 +1,21 @@
...
-    <div className={$layout.container}>
-      <NavBar />
-      <main className={$layout.main}>{children}</main>
-    </div>
+    <>
+      <div className={$layout.container}>
+        <NavBar />
+        <main className={$layout.main}>{children}</main>
+      </div>
+    </>
```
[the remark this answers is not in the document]
tessathornberry: Yes, good idea - I've done away with the fragment.

src/pages/settings.tsx
tessathornberry: Since each user should only have access to their own settings, and it is currently simply a skeleton/plceholder file, I am happy to simplify this for now and let adding any future potential protections for a future PR, once we have some settings to put in there. ;)

src/pages/signInPage.tsx
tessathornberry: You got it, thank you!

tessathornberry:
This safely redirects an unautheticated user to the sign-in page from any non-public URL.
A future issue will be to disallow any "loading" messages or wrapper-based queries, which are currently bouncing back as 401's to GET requests.

codetheweb (approved):

jonahchoi (approved):
Looks good. Thanks Tessa!

PR #175, opened 2023-08-10
Fixed empty person table error

ucheoh:


jonahchoi:
What error is this fix for? I believe the tRPC call already returns an empty list, and I don't see any errors on my end (on main branch).

ucheoh:
> What error is this fix for? I believe the tRPC call already returns an empty list, and I don't see any errors on my end (on main branch).

They are in reference to these errors that popped up in the browser when navigating to the "/students" page logged out.

screen_shot_2023-07-31_at_1 46 18_pm
screen_shot_2023-07-31_at_1 42 02_pm

jonahchoi (commented):
This is more an issue of permissions, since unauthenticated users shouldn't be able to access this page in the first place. (Also probably a sign we need better error handling). I'm not too sure if these changes are necessary in that case. Let's wait until the new protected pages changes come in and see if this is still needed.

Issue #158, opened 2023-07-25
Restructure components to group by component [Eng]


jonahchoi:
Remove case_manager and other folders, and reorganize component file structure to group by component.

Ex:
-  components/
    - table/
        - table.tsx
        - table.module.css
    - button/
        - button.tsx
        - button.module.css

BeeSeeWhy:
completed and merged to main

PR #186, opened 2023-08-22
restructured components folder

BeeSeeWhy:
issue #158


src/components/personTable/PersonTable.tsx
jonahchoi, line 1: This file can be with the rest of the table files. We can rename this one to just 'table.tsx'.

BeeSeeWhy: I tried renaming to table.tsx and moving to table folder, but it kept giving me an error saying that it couldn't find "@/styles/Button.module.css" in the imports. If I left as personTable, everything was ok.

jonahchoi: I think the issue is with table.ts having the same name. For now, can you rename `table.ts` -> `tableTypes.ts`, and `PersonalTable.tsx` -> `table.tsx`? We might end up removing that types file in the future anyway.

jonahchoi (approved):
Looks good

PR #195, opened 2023-09-01
enabled single click for file upload

shin101:


shin101:
https://github.com/sfbrigade/compass/assets/44033225/3a692c9d-c114-44f7-b4a2-b381a89cbdb2

This is what it looks like now

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -111,27 +113,23 @@ const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
   }, [uploadSuccess]);
 
   return (
-    <div className={styles.uploadContainer}>
+    <div className={styles.uploadContainer} onClick={onClickStartButton}>
```
jonahchoi, line 116: Anything that functions as a button should probably use `button` tags. But should this onclick be on the container? Shouldn't it be on the div on line 118?

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -111,27 +113,23 @@ const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
...
-            className={styles.actionButton}
+            style={{ display: "none" }}
           />
+          <CameraIcon />
+          {title}
+        </div>
+      )}
+      {currentStep === UploadStep.TakePicture && (
+        <>
           <div>
```
jonahchoi, line 133: The fragment and extra div don't look necessary.

jonahchoi (changes requested):

src/components/uploadPicture/uploadImage.tsx
shin101: will do. thanks for the feedback!

shin101:
Made updates based on feedback :)

jonahchoi (approved):

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -113,34 +115,23 @@ const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
...
-        <>
+        <button className={styles.actionButton} onClick={onClickStartButton}>
           <input
             ref={fileInputRef}
             type="file"
             accept="image/*;capture=camera"
             capture="environment"
             onChange={handleFile}
-            className={styles.actionButton}
+            style={{ display: "none" }}
```
codetheweb, line 125: a little late now but wondering if it'd be possible to style the input directly instead of wrapping with a button?

jonahchoi: Input type files tend to not like styling. It would be slightly better to use a label instead, making the onClick function unecessary, but the principle remains the same. https://stackoverflow.com/a/25825731

Issue #152, opened 2023-07-18
CM flow: Create student's individual pages [Eng]


jonahchoi:
- [ ] Create tabs for each category (goals, staff, etc.)
- [ ] Render goals in goals tab
- [ ] On click, dropdown should show benchmarks(subgoals) relating to goal
- [ ] Each benchmark should show basic info, star/favorite button, and view more button
- [ ] Button to add new goals (probably a modal, tbd)

![Image](https://github.com/sfbrigade/compass/assets/104597125/e6881634-e237-4a56-a442-cebd707d8f77)

![Image](https://github.com/sfbrigade/compass/assets/104597125/69a34f31-fe04-420d-aa90-0eac561cf761)

jonahchoi:
Blocked by goals/subgoals/tasks backend

jonahchoi:
Unblocked now that backend is merged

KCCPMG:
Most recent draft of student page (still pending):



![Image](https://github.com/sfbrigade/compass/assets/39074889/04eebb4d-99b2-423a-80c8-0e0d3c2c9a5f)

PR #196, opened 2023-09-06
Show Student's Page with details

hieungo89:
Referencing issue #152 - Adding Texture to Student's page

Things included in this PR:

- All students page
    - [x] Added `Grade` & `IEP end date`
    - [x] Students with no IEP will have a 'None' under `IEP end date`
- Individual Student's Page
    - [x] Moved `Iep.tsx` from a page to component 
    - [x] Shows `No IEP Default image` when no IEP is active
    - [x] Shows `No Goals Default image` when no goals are in DB
    - [x] Reconfigured CSS for student's page according to designs
    - [x] Added Modal for `creating student IEP` & `archiving student`
- Backend
    - [x] Added `getMyStudentsAndIepInfo` to `case_manager.ts`
    - [x] Added `getActiveStudentIep` to `student.ts`
    - [x] Added `getStudentDetailById` to `student.ts`, not yet used
    - [x] Added `is_active` to the database `iep table` 

Here are some photos of the finished product:
Students page
Individual Student Page without IEP
Individual Student Page without goals
Individual Student Page with goals


src/pages/students/[student_id].tsx
```diff
@@ -54,80 +82,203 @@ const ViewStudentPage = () => {
...
+                padding: "1rem",
+              }}
+            >
+              <div>IEP End Date:</div>{" "}
+              <div style={{ alignSelf: "center" }}>
+                {activeIep ? (
+                  <>{new Date(activeIep.end_date ?? "").toLocaleDateString()}</>
+                ) : (
+                  <>None</>
+                )}
```
Comment (jonahchoi) [inline src/pages/students/[student_id].tsx:126]: 🐧 Don't think the fragments are necessary

src/components/subgoal/Subgoal.tsx
```diff
@@ -20,23 +22,46 @@ const Subgoals = ({ subgoal }: SubgoalProps) => {
   //   });
   // }
   const assignToPara = () => {
-    task.mutate({
-      subgoal_id: subgoal.subgoal_id,
-      due_date: new Date(2023, 8, 20),
-      trial_count: 5,
-    });
+    // task.mutate({
```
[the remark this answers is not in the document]
hieungo89: I believe we didn't dive into any goals or subgoals files. This could probably be the result of rebase merging.

src/backend/routers/student.ts
```diff
@@ -35,6 +35,74 @@ export const student = router({
...
+          "subgoal.subgoal_id",
+          "subgoal.description as subgoal_description",
+          "instructions",
+          "target_max_attempts",
+        ])
+        .execute();
+
+      // return rows as an array of objects which in turn contain objects representing tables
+      // [{{student, iep, goal, subgoal}}, {{student, iep, goal, subgoal}}, ...]
+      const tableRows = rows.map(function (row) {
```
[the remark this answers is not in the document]
KCCPMG: I was actually planning on doing just that, especially after seeing your implementation of that on the Iep procedure that you wrote. As of right now, this procedure isn't called on the front end, so it's something of a work in progress that may be abandoned all together. To that point, when we were discussing this yesterday, it opened up the larger question of how we should proceed with this kind of deeply nested document where everything is appearing on one page. Should we make this nested request, keep everything in the parent state, and pass everything down as props? Would this also mean needing to set up context or passing down additional procedures as props? Or should we expect to continue with the current approach of every rendered component calling its own procedure to get its own data? We were going to discuss this in the meeting as well but I think it's good to give you a head start if you'd like to think about it advance, plus writing this out helped me organize our thoughts about it.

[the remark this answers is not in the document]
jonahchoi: I'm not an expert, so this is just my opinion. I don't think every component should make an API call, because one student page could potentially be making tens or hundreds of calls (multiple subgoals for multiple goals for multiple* ieps). It may not be necessary to set up context if the props are simple, but it depends on implementation and can definitely be useful. We should definitely have this conversation with the team.

src/backend/routers/case_manager.ts
```diff
@@ -17,6 +17,28 @@ export const case_manager = router({
     return result;
   }),
 
+  getMyStudentsAndIepInfo: authenticatedProcedure.query(async (req) => {
+    const { userId } = req.ctx.auth;
+
+    const result = await req.ctx.db
+      .selectFrom("student")
+      .fullJoin("iep", "iep.student_id", "student.student_id")
```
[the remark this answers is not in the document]
hieungo89: I redefined the query in the backend to accept return data for each student if 

1. student does not have any ieps
2. student has an active iep
3. student has no active iep, but did have past ieps

src/components/subgoal/Subgoal.tsx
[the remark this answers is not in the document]
hieungo89: I believe this is commented out and still in the code base.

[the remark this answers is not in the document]
jonahchoi: Why is it commented out?

src/components/table/tableTypes.ts
```diff
@@ -17,6 +19,12 @@ export interface ParaHeadCell extends HeadCell {
...
 
+export interface IepHeadCell extends HeadCell {
+  id: keyof Iep;
+}
+
+export type StudentWithIepHeadcell = StudentHeadCell | IepHeadCell;
+
 export function isStudent(person: Student | Para): person is Student {
   return (person as Student).student_id !== undefined;
 }
```
jonahchoi, line 30: Since we're using this function to test for StudentWithIep, we should convert it to test for that specifically. 
```suggestion
export function isStudentWithIep(person: StudentWithIep | Para): person is StudentWithIep {
  return (person as StudentWithIep).student_id !== undefined && (person as StudentWithIep).iep_id !== undefined;
}
```

src/components/subgoal/Subgoal.tsx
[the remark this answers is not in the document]
hieungo89: Not sure. Could be an error on our part when we rebase from main. I can change it back.

src/backend/routers/student.ts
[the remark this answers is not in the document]
KCCPMG: As per our conversation in the meeting, especially with @amantri, this procedure has been removed, and I'm marking this as resolved.

src/backend/routers/case_manager.ts
```diff
@@ -17,6 +17,44 @@ export const case_manager = router({
...
 
+  getMyStudentsAndIepInfo: authenticatedProcedure.query(async (req) => {
+    const { userId } = req.ctx.auth;
+
+    const studentData = await req.ctx.db
+      .selectFrom("student")
+      .where("assigned_case_manager_id", "=", userId)
+      .fullJoin("iep", (join) =>
+        join.onRef("iep.student_id", "=", "student.student_id")
+      )
```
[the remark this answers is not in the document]
KCCPMG: In light of this week's discussions, at this point it sounds to us like the best course of action at this point is to simplify this to just make it a join of iep and student on student_id, returning all where the student's assigned case manager is the user. This means not worrying about start or end dates or grabbing just one iep per student since we're operating on the assumption that there's only one iep per student. We're expecting to have a commit today that will have the revised procedure.

hieungo89:
@codetheweb & @jonahchoi, please verify. We had some trouble with rebasing. Not sure if the current push affects your work.

hieungo89:
Also, Not sure why the PR checks are not working correctly...

codetheweb:
> @codetheweb & @jonahchoi, please verify. We had some trouble with rebasing. Not sure if the current push affects your work.

I went ahead and merged main into this branch which fixed it. Looks like there's some linting errors that need to be resolved now.

src/backend/routers/student.ts
```diff
@@ -86,5 +86,25 @@ export const student = router({
...
+      const { student_id } = req.input;
+
+      const result = await req.ctx.db
+        .selectFrom("iep")
+        .where("student_id", "=", student_id)
+        .where("is_active", "=", true)
+        .selectAll()
+        .executeTakeFirst();
+
+      return result || null;
```
codetheweb, line 106: type of `result` is already `Iep | null`

```suggestion
      return result;
```

src/backend/routers/student.ts
```diff
@@ -86,5 +86,25 @@ export const student = router({
       return result;
     }),
 
+  // TODO: get only the active IEP from the student
```
codetheweb, line 89: it looks like this condition currently holds?

src/backend/routers/case_manager.ts
```diff
@@ -28,10 +51,11 @@ export const case_manager = router({
         first_name: z.string(),
         last_name: z.string(),
         email: z.string().email(),
+        grade: z.string(),
```
codetheweb, line 54:
```suggestion
        grade: z.number(),
```

(then don't need the `Number()` cast below)

src/backend/routers/student.ts
hieungo89: I think the reason for the return with null is that if the student has no iep data, then `result` will be `undefined`.

src/backend/routers/case_manager.ts
jonahchoi: I think the issue here is that the grade is input as a text, so we would need to convert it to Number either here or on the frontend. Unless setting z.number() automatically converts strings to numbers?

src/backend/routers/student.ts
codetheweb: oh hmm my bad, thought kysely returned null in this case

src/backend/routers/case_manager.ts
codetheweb: I would recommend converting on the frontend. Alternatively you can do `z.coerce.number()` to automatically parse strings into numbers.

src/components/layout/Layout.module.css
```diff
@@ -7,7 +7,7 @@
 .main {
   flex: 10;
   position: relative;
-  background-color: var(--background);
+  background-color: var(--grey-80);
```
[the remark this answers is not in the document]
jonahchoi: Wanted to circle back on this one. Are we changing all background to be grey-80?

src/backend/routers/student.ts
hieungo89: Actually, you are right @codetheweb. I think the problem that it's not returning null is that line 104 is supposed to be `.executeTakeFirstOrThrow()`. I'm fixing this issue in the upcoming push.

src/backend/routers/student.ts
hieungo89: I believe this condition will hold for our MVP since we only need to retrieve one student iep.

src/backend/routers/case_manager.ts
hieungo89: We converted all grades to numbers and will only accept numbers in the frontend.

src/components/layout/Layout.module.css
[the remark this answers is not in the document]
hieungo89: I see what you meant. I can fix it to show only on the student page.

codetheweb (approved):

jonahchoi (approved):

Issue #182, opened 2023-08-18
Fix notes not saving when updating a counter [Eng]


jonahchoi:
The code can be found in `src/pages/benchmarks/[benchmark_id]/index.tsx`. To prevent too many API calls, the notes will mutate the database 500ms after typing has stopped. But if the user clicks the counter before that, a new task will be returned, which updates the `notesValue` state, effectively reverting the notes back to its previous value. Find a solution to keep the new text without requiring multiple updates.

PR #197, opened 2023-09-07
Fixed notes bug and added auto save features

jonahchoi:
#182 - All user inputs use the same debounced function to prevent excess API calls and ensure data is uploaded correctly. Added a new indicator to show if data has been saved and a custom hook to popup a warning if the user tries to leave the page with unsaved data.


src/pages/benchmarks/[benchmark_id]/index.tsx
```diff
@@ -47,27 +48,42 @@ const BenchmarkPage = () => {
     onSuccess: async () => await utils.iep.getSubgoalAndTrialData.invalidate(),
   });
 
-  const [notesValue, setNotesValue] = useState("");
+  const [notesInputValue, setNotesInputValue] = useState("");
+  const [successInputValue, setSuccessInputValue] = useState(0);
+  const [unsuccessInputValue, setUnsuccessInputValue] = useState(0);
+  const [isInputChange, setIsInputChange] = useState(false);
```
Comment (codetheweb) [inline src/pages/benchmarks/[benchmark_id]/index.tsx:54]: I think this could be a computed property based on something like `const hasInputChanged = notesInputValue !== (currentTrial?.notes ?? "") || // ... other properties`?

src/pages/benchmarks/[benchmark_id]/index.tsx
```diff
@@ -95,20 +111,41 @@ const BenchmarkPage = () => {
...
+    [notesInputValue, successInputValue, unsuccessInputValue]
   );
+
+  // BUG?: Sometimes if the user reloads/navigates away and confirms, the update has time to go through and data is saved. Is this something we should fix?
+  const checkUnsavedData = useCallback(() => {
+    if (
+      currentTrial?.notes !== notesInputValue ||
+      currentTrial?.success !== successInputValue ||
+      currentTrial?.unsuccess !== unsuccessInputValue
+    ) {
```
Comment (codetheweb) [inline src/pages/benchmarks/[benchmark_id]/index.tsx:141]: could use `isInputChange` here?

codetheweb (approved):

src/pages/benchmarks/[benchmark_id]/index.tsx
Comment (jonahchoi) [reply src/pages/benchmarks/[benchmark_id]/index.tsx:54]: At first I didn't think this would work, since the initial state would be different from currentTrial. But after testing, I realized the debounce will wait until after the setState, giving it enough time to not run the unnecessary update. Good catch!

src/pages/benchmarks/[benchmark_id]/index.tsx
```diff
@@ -95,20 +115,35 @@ const BenchmarkPage = () => {
...
-    [notesValue]
+    [notesInputValue, successInputValue, unsuccessInputValue]
   );
+
+  // BUG?: Sometimes if the user reloads/navigates away and confirms, the update has time to go through and data is saved. Is this something we should fix?
+  const checkUnsavedData = useCallback(() => {
+    if (hasInputChanged) {
+      return true;
+    }
+    return false;
```
Comment (codetheweb) [inline src/pages/benchmarks/[benchmark_id]/index.tsx:142]:
```suggestion
    return hasInputChanged
```

codetheweb (approved):

src/pages/benchmarks/[benchmark_id]/index.tsx
Comment (jonahchoi) [reply src/pages/benchmarks/[benchmark_id]/index.tsx:142]: LOL 🤦‍♂️

Comment (jonahchoi) [reply src/pages/benchmarks/[benchmark_id]/index.tsx:142]: Actually I'll do you one better. We don't even need this callback since we can just pass hasInputChanged in the useConfirmBeforeLeave hook.

Issue #198, opened 2023-09-19
Update /benchmarks to load benchmarks instead of tasks [ENG]


tessathornberry:
Currently, the /benchmarks tab loads all task instances, when it should perhaps load a list of benchmarks, and clicking on the benchmark should load the task instances, for assignment to staff or data collection

![Image](https://user-images.githubusercontent.com/105292208/268826738-b46b1d71-ee5b-41da-909d-ffbb8901a4f9.png)

jonahchoi:
Seems like the issue is actually being able to press the 'assign' button multiple times. Each click adds a brand new task, resulting in the repeats seen here. I haven't been able to recreate it in other ways. Let me know if you see anything different. Since the UI hasn't been made for assigning benchmarks, I could add a quick change for the temporary function we have to prevent multiple assignments.

PR #204, opened 2023-09-21
Quick bug fix to prevent multiple assignments of benchmarks.

jonahchoi:
#198

### DIFF
--- a/src/backend/routers/iep.ts
+++ b/src/backend/routers/iep.ts
@@ -92,6 +92,23 @@ export const iep = router({
     .mutation(async (req) => {
       const { subgoal_id, due_date, trial_count } = req.input;
       const { userId } = req.ctx.auth;
+
+      const shouldAdd = await req.ctx.db
+        .selectFrom("task")
+        .selectAll()
+        .where((eb) =>
+          eb.and([
+            eb("subgoal_id", "=", subgoal_id),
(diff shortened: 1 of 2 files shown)


jonahchoi:
Just so you know: this fix is only for our temporary assignment function, and will likely be overwritten later on.

src/backend/routers/iep.ts
```diff
@@ -92,6 +92,23 @@ export const iep = router({
...
       const { subgoal_id, due_date, trial_count } = req.input;
       const { userId } = req.ctx.auth;
+
+      const shouldAdd = await req.ctx.db
+        .selectFrom("task")
+        .selectAll()
+        .where((eb) =>
+          eb.and([
+            eb("subgoal_id", "=", subgoal_id),
+            eb("assignee_id", "=", userId),
```
codetheweb, line 102: should we have a database constraint on this to prevent duplicates?

codetheweb (approved):

src/backend/routers/iep.ts
jonahchoi: I think there might be a case where we allow duplicates in certain situations. Like if we want to redo the same benchmark or something. Let's hold off on that until we get the frontend designs and discuss that with the design team.

Issue #206, opened 2023-09-24
Move MyStudents/MyParas to pages


jonahchoi:
Both MyStudents and MyParas are currently components that don't get reused and are the only things rendered on their respective pages. I think they would fit more in the 'pages' category.

PR #208, opened 2023-09-24
Moved around some files to more appropriate places.

jonahchoi:
#206 - Moved MyStudents/MyParas components into students/staff pages. Also moved `tableTypes` to global types file.


src/pages/staff/index.tsx
```diff
@@ -1,8 +1,71 @@
...
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    const data = new FormData(event.currentTarget);
+
+    createPara.mutate({
+      first_name: data.get("first_name") as string,
+      last_name: data.get("last_name") as string,
+      email: data.get("email") as string,
+    });
```
codetheweb, line 34: I think this is old code but would recommend
```suggestion
    await createPara.mutateAsync({
      first_name: data.get("first_name") as string,
      last_name: data.get("last_name") as string,
      email: data.get("email") as string,
    });
```
so the form doesn't get reset if there's an error

src/types/global.ts
```diff
@@ -16,3 +16,36 @@ export interface ParaTaskCard {
...
+
+export interface ParaHeadCell extends HeadCell {
+  id: keyof Para;
+}
+
+export interface IepHeadCell extends HeadCell {
+  id: keyof Iep;
+}
+
+export type StudentWithIepHeadcell = StudentHeadCell | IepHeadCell;
```
codetheweb, line 42: 🐧 I would probably keep these in `table/table.tsx` since they're specific to that component

codetheweb (approved):

src/types/global.ts
jonahchoi: Some of these types are being used in the `student/index` and `staff/index` pages. Would you keep the base `HeadCell` in this file, and move types that are only used in one file to that file?

codetheweb: it's generally best to keep the type close to the thing that "owns" it--by ownership I mean the thing that takes parameters of the given type and results in terminal behavior

in this case the table component owns the cell types

components that want to render a table can import the cell types from the table component

another way to look at it is what types you would have to change if you want to refactor a particular component

codetheweb (approved):

Issue #95, opened 2023-06-01
Para flow: Attach images to a trial data collection [Eng]


amantri:
Upload the images to the cloud and associate it with the trial data collection.

Figma: https://www.figma.com/file/m09znscRXNqSziAiEbNLTf/Compass-Wireframes?type=design&node-id=4197-70166&t=duBg5z5djAaQ44DF-4

Mocks:

<img src="https://github.com/sfbrigade/compass/assets/829820/159304e0-65bc-41ec-aa0f-f251129f4f94" width=40%>

MarkOfosu:
This looks interesting to me and so I am happy to start working on it. Hopefully, I can make my first pull request soon!!

codetheweb:
The resulting React component should use the `.getPresignedUrlForFileUpload()` and `.finishfileUpload()` procedures to upload files to the backend (see `src/backend/routers/files.test.ts` for example usage). The component API should look something like:

```tsx
<FileUpload fileIds={[]} onChange={() => { ... }}/>
```

where

```ts
interface FileUploadOptions {
  fileIds: string[]
  onChange: (newFileIds: string[]) => void
}
```

MarkOfosu:
is the take a picture of students' work button in the image that came with the ticket the same as the file upload component that is shown above?

codetheweb:
Yep, you'll probably want to add another prop to the above example component to allow customizing the title.

MarkOfosu:
Also, at what point are the images actually uploaded? Is it right after the user selects the image?

codetheweb:
It's up to you. I would:

- open the file picker / camera after clicking the button
- when control is given back to our application and something has been selected, start the upload, disable the button, and add an activity indicator
- once the upload is finished change the state of the button to indicate that the file has been uploaded (maybe a checkmark and/or show the file name?)

MarkOfosu:
Thanks Max!

PR #209, opened 2023-09-25
Add file picker to data collection page

codetheweb:
Closes https://github.com/sfbrigade/compass/issues/95.

Demo:

https://github.com/sfbrigade/compass/assets/7410405/e814f480-bb0f-449f-bbe3-e71afa4501f7


src/components/uploaded-file/uploaded-file.tsx
```diff
@@ -0,0 +1,61 @@
...
+  useEffect(() => {
+    mutate({ file_id: fileId });
+  }, [mutate, fileId]);
+
+  return (
+    <div className={styles.container}>
+      {data ? (
+        <>
+          {data.content_type.startsWith("image/") ? (
+            <img src={data.url} className={styles.image} />
```
hieungo89, line 26: Does the `` need an alt tag?

Also, is there a difference in using `` here instead of `<Image>` from next?

src/backend/db/migrations/1_initial-migrations.sql
```diff
@@ -108,6 +108,14 @@ CREATE TABLE "trial_data" (
   created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE "trial_data_file" (
+  trial_file_id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+  trial_data_id UUID REFERENCES "trial_data" (trial_data_id),
```
jonahchoi, line 113: Couldn't the `file` table directly reference `trial_data`? Not sure if there's other places we would upload files.

src/backend/routers/iep.ts
```diff
@@ -271,6 +271,18 @@ export const iep = router({
                 "trial_data.notes",
                 "trial_data.created_at",
               ])
+              .select((eb) =>
```
jonahchoi, line 274: 🐧 Above, there's one `select` that includes columns and the `jsonArrayFrom`. Here, it's separated into two `select`s. Looks like either way works, but for the sake of consistency/readability, I think we should stick with one way of writing these.

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -19,14 +18,10 @@ enum UploadStep {
   UploadPicture,
 }
 
-const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
-  title,
-  fileName,
-  onUpload,
-}) => {
+const UploadImage: React.FC<UploadImageProps> = ({ label, onUpload }) => {
```
jonahchoi, line 21:
```suggestion
const UploadImage = ({ label, onUpload }: UploadImageProps) => {
```

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -39,31 +34,23 @@ const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
...
-    setCapturedImage(null);
+    setSelectedFile(null);
   };
 
   const onClickUploadPicture = () => {
     setCurrentStep(UploadStep.Start);
-    setCapturedImage(null);
+    setSelectedFile(null);
     setUploadSuccess(true);
     setUploadSuccess(false);
```
jonahchoi, line 51: Not your code, but do you know why this is setting to true then false immediately?

src/components/uploaded-file/index.ts
jonahchoi, line 1: 🐧 All our component files/folders are camelCased.

src/components/uploadPicture/uploadImage.tsx
```diff
@@ -135,9 +117,9 @@ const UploadImageComponent: React.FC<UploadImageComponentProps> = ({
       )}
       {currentStep === UploadStep.UploadPicture && (
         <div className={styles.uploadContainer}>
-          {capturedImage && (
+          {selectedFile && (
             <svg>
-              <image href={URL.createObjectURL(capturedImage)} />
+              <image href={URL.createObjectURL(selectedFile)} />
             </svg>
```
jonahchoi, line 123: Is there a reason this has to be an svg instead of just an Image tag? Not really sure if there's a reason to change it though.

src/components/uploaded-file/uploaded-file.tsx
jonahchoi: Alt could just be "User-uploaded image". Also the next image uses some optimizations to make things faster (can't say exactly what though).

src/pages/benchmarks/[benchmark_id]/index.tsx
```diff
@@ -137,6 +150,28 @@ const BenchmarkPage = () => {
...
+      trial_data_id: currentTrial?.trial_data_id,
+    });
+  };
+
+  const handleFileDelete = async (fileId: string) => {
+    if (!currentTrial) {
+      throw new Error("Trial data has not yet loaded");
+    }
+
+    await removeFileFromTrialDataMutation.mutateAsync({
```
Comment (jonahchoi) [inline src/pages/benchmarks/[benchmark_id]/index.tsx:169]: Correct me if I'm wrong, but looks like we're removing the association but not actually deleting the uploaded file, which would be concerning to me as a user. If so, we should update this to delete the file as well.

src/components/uploaded-file/uploaded-file.tsx
codetheweb: sure, will add an alt

> Also, is there a difference in using  here instead of <Image> from next?

in our case using `next/image` would likely make things slower as it can take 1-5s upon first load to transform and cache the result (especially because we're not using Vercel's optimized transform pipeline)

it's useful for public images or ones that are shown to many users, but I think just using `` is likely the better experience in this case

codetheweb: 1de9bc2

src/backend/db/migrations/1_initial-migrations.sql
codetheweb: we could, but imo this adds very complexity 🤷 

some places where we might also use `file` in the future:

- custom user avatars
- uploaded audio notes

src/backend/routers/iep.ts
codetheweb: 62f4a46

src/components/uploadPicture/uploadImage.tsx
codetheweb: not sure, I'll clean up this file a little bit

codetheweb: 0825676

src/components/uploaded-file/index.ts
codetheweb: 0825676

src/components/uploadPicture/uploadImage.tsx
codetheweb: 0825676

src/pages/benchmarks/[benchmark_id]/index.tsx
Comment (codetheweb) [reply src/pages/benchmarks/[benchmark_id]/index.tsx]: 88a3f23

jonahchoi (approved):
Looks good

Issue #201, opened 2023-09-19
Correct Authenticated-Page to disallow GET/Trpc calls [ENG]


tessathornberry:
Add a conditional on authenticated-page to disallow GET and Trpc calls even with 401 responses.

![Image](https://user-images.githubusercontent.com/105292208/268829176-5ef54865-ab6c-4d48-9580-8da133063969.png)

PR #211, opened 2023-10-12
Redirect user if missing auth cookie

codetheweb:
Closes https://github.com/sfbrigade/compass/issues/201.

Not perfect, but better than nothing.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,0 +1,21 @@
+import { NextRequest, NextResponse } from "next/server";
+
+/**
+ * Our API routes are already protected, this middleware applies to UI routes like /students.
+ * If a user isn't logged in, we redirect them to sign-in.
+ * This **does not** check for the validity of the session, only that a cookie exists; so it shouldn't be generally used to protect routes.
+ *
+ * Ideally we would use NextAuth's built-in Next.js middleware: https://next-auth.js.org/configuration/nextjs#middleware
+ * However, their middleware doesn't yet work with database sessions.
+ */
+export function middleware(request: NextRequest) {
+  if (!request.cookies.get("next-auth.session-token")) {
+    const url = request.nextUrl.clone();
+    url.pathname = "/signInPage";
+    return NextResponse.redirect(url);
+  }
+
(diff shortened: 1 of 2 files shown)


src/middleware.ts
```diff
@@ -0,0 +1,21 @@
+import { NextRequest, NextResponse } from "next/server";
+
+/**
+ * Our API routes are already protected, this middleware applies to UI routes like /students.
+ * If a user isn't logged in, we redirect them to sign-in.
+ * This **does not** check for the validity of the session, only that a cookie exists; so it shouldn't be generally used to protect routes.
+ *
+ * Ideally we would use NextAuth's built-in Next.js middleware: https://next-auth.js.org/configuration/nextjs#middleware
+ * However, their middleware doesn't yet work with database sessions.
```
tessathornberry, line 9: Thank you for determining this, @codetheweb .

tessathornberry, line 9: Thank you for determining this, @codetheweb .

src/middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function middleware(request: NextRequest) {
+  if (!request.cookies.get("next-auth.session-token")) {
+    const url = request.nextUrl.clone();
```
tessathornberry, line 13: How you put these properties together really shows the thoughtfulness you use when coding.

src/middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+    url.pathname = "/signInPage";
+    return NextResponse.redirect(url);
+  }
+
+  return NextResponse.next();
+}
+
+export const config = { matcher: ["/((?!img|_next|api|signInPage).*)"] };
```
tessathornberry, line 21: We can add the "About" page to this when we have one. :) Great work!

tessathornberry (approved):
@codetheweb This works great - I agree with you that is imperfect, but it works for what we need it for and you handled it masterfully - I can see that it really required someone with your level of skill to accomplish this, and you've done a great job - thank you!

PR #224, opened 2023-10-17
About page

SeldomseenSchweig:
this page describes the app and it's purpose. It is gotten to through the Compass logo


src/client/lib/authenticated-page.tsx
```diff
@@ -10,7 +10,7 @@ export const requiresLogin =
     const { status } = useSession();
 
     // Include any public pathnames here
-    const publicPaths = ["/signInPage"];
+    const publicPaths = ["/signInPage", "/about"];
```
tessathornberry, line 13: Good job here. Since the app has been updated to include middleware, in "Middleware.ts", please pull down and merge that code with this, and then add it to the matcher in the RegEx in the file so it's not a blocker for those who are not signed in.

src/components/navbar/NavBar.tsx
```diff
@@ -46,6 +47,11 @@ const NavBar = () => {
               <p className={$navbar.linkTitle}>Settings</p>
             </Link>
             <br />
+            <Link href="/about" className={$navbar.link}>
+              <Info className={$navbar.icon} />
```
tessathornberry, line 51: Please be sure you are adding the property types to the appropriate .css file - and fix some of the "any" type errors.

src/pages/about.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <>
+      <h2>About Compass</h2>
+
+      <div>
+        Compass is a civic app we are building for the San Francisco Unified
+        School District. The app seeks to assist Teachers with setting and
+        tracking goals for students with disabilities.
+      </div>
+      <div></div>
+    </>
```
tessathornberry, line 15: Is there a reason we have the fragments instead of divs? (It makes it easier to apply CSS if they are specific elements rather than fragments.)
Minor cleanup request - please remove extra code we are not currently using, like the imports and empty div - we can add needed imports and components later.
Good job - this is a good starting point.

tessathornberry (changes requested):

tessathornberry:
We can add more functionality and text to this, but since I have not heard from either of you I will go ahead and merge this. Thanks for working on it.

Issue #214, opened 2023-10-16
Autofill IEP end date [ENG]


tessathornberry:
Currently the IEP date asks for a beginning and end date. It would make much more sense to have the end date autofill with the date in one-year-minus-one-day .

PR #225, opened 2023-10-19
Used date-fns to autofill IEP end date and correctly format IEP dates in student_id

tessathornberry:
#214 
Used date-fns to create auto-fill end dates once user puts a start date in, and also format dates in student_id  so that IEP date values are as per user selection both in the front end and in the database.

* Date values from ```<input type="date" />``` come out in format "YYYY-MM-DD", which formats to UTC time rather than local time when passed into new Date(), which was putting the IEP dates off by 1 day too early.


codetheweb (approved):
Nice

PR #226, opened 2023-10-26
Fix staging login redirect loop

codetheweb:
Apparently next auth uses a different cookie name for HTTPS origins.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -9,7 +9,11 @@ import { NextRequest, NextResponse } from "next/server";
  * However, their middleware doesn't yet work with database sessions.
  */
 export function middleware(request: NextRequest) {
-  if (!request.cookies.get("next-auth.session-token")) {
+  const sessionToken =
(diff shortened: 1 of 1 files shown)


src/middleware.ts
```diff
@@ -9,7 +9,11 @@ import { NextRequest, NextResponse } from "next/server";
  * However, their middleware doesn't yet work with database sessions.
  */
 export function middleware(request: NextRequest) {
-  if (!request.cookies.get("next-auth.session-token")) {
+  const sessionToken =
+    request.cookies.get("next-auth.session-token") ??
+    request.cookies.get("__Secure-next-auth.session-token");
```
tessathornberry, line 14: All the fascinating intricacies of various frameworks.
Thanks for fixing this, @codetheweb

tessathornberry (approved):
We are getting an "API handler should not return a value, received object." in the terminal console, but that is from the previous PR.

PR #229, opened 2023-10-31
Fix getActiveStudentIep frontend query to stop calling if call failed

hieungo89:
Before: 
Screenshot 2023-10-30 at 5 03 40 PM

After: 
Screenshot 2023-10-30 at 5 02 05 PM


codetheweb (approved):
nice

PR #233, opened 2023-11-03
BUG: Empty staff list shows student's empty state #232

hieungo89:
Referencing issue #232. I adjusted the specification of the empty state so Staff page shouldn't be affected by Student's list empty state anymore.

**Before:**
Screenshot 2023-11-03 at 10 11 17 AM

**After**:
Screenshot 2023-11-03 at 10 09 31 AM

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -135,7 +135,7 @@ function EnhancedTableToolbar({
 }: EnhancedTableToolbarProps) {
   return (
     <>
-      {totalRows === 0 ? (
+      {totalRows === 0 && type === "Students" ? (
         <h2 className={$table.tableTitle}>{type}</h2>
       ) : (
         <Toolbar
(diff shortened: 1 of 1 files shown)


codetheweb (approved):
should we edit the copy for the empty staff state instead?

hieungo89:
> should we edit the copy for the empty staff state instead?

Niko confirmed that there won't be an empty staff state since the staff page will always have at least the current signed-in staff and never be empty.

tessathornberry (approved):
Looks great, Hieu! Nice work! Maybe we can make a separate issue to add the user to the staff list, too! What do you think?

hieungo89:
> Looks great, Hieu! Nice work! Maybe we can make a separate issue to add the user to the staff list, too! What do you think?

I'm planning to work on that next. When staff sign in for the first time, their name is automatically added to the staff list so the list won't be empty.

PR #242, opened 2023-11-11
Assign self to staff list

hieungo89:
Add self to staff list & disable `unassign Para` when viewing self.

- **All Staff page View:**
Screenshot 2023-11-13 at 7 33 58 PM

- **View of Self:**
Screenshot 2023-11-10 at 5 47 44 PM

- **Original - Individual para page view:**
Screenshot 2023-11-10 at 5 47 53 PM

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -378,7 +378,6 @@ export default function EnhancedTable<
               )}
               {visibleRows.map((row) => {
                 const labelId = row.email;
-                console.log("row ", row);
 
                 return (
                   <StyledTableRow
@@ -398,7 +397,7 @@ export default function EnhancedTable<
                     <TableCell component="th" id={labelId} scope="row">
                       {row.first_name}
                       {!isStudentWithIep(row) &&
-                        row.role === "admin" &&
+                        Object.keys(row).length < 9 &&
                         " (me)"}
                     </TableCell>
                     <TableCell align={"left"}>{row.last_name}</TableCell>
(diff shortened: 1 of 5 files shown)


codetheweb (approved):
lgtm other than CI checks

src/components/table/table.tsx
```diff
@@ -396,6 +396,9 @@ export default function EnhancedTable<
                   >
                     <TableCell component="th" id={labelId} scope="row">
                       {row.first_name}
+                      {!isStudentWithIep(row) &&
+                        Object.keys(row).length < 9 &&
+                        " (me)"}
```
tessathornberry, line 401: Would you please document in a comment that this is a conditional based on the length of the para row vs to case manager row, since that is the only distinction?

src/pages/staff/[user_id].tsx
```diff
@@ -42,18 +43,20 @@ const ViewParaPage = () => {
...
+        </button>
+      )}
 
       {unassignParaPrompt ? (
         <div>
           <p>
-            Are you sure you want to unassign {para?.first_name}{" "}
-            {para?.last_name}?
+            Are you sure you want to unassign
+            {para?.first_name} {para?.last_name}?
```
Comment (tessathornberry) [inline src/pages/staff/[user_id].tsx:59]: This will have to be done with template literals inside of backticks if you want the space to be recognized without the ```{" "}```

tessathornberry (changes requested):
Great work integrating the two PRs. Two small change requests from me.

src/components/table/table.tsx
hieungo89: Screenshot 2023-11-14 at 5 11 25 PM

tessathornberry (approved):
Good work, @hieungo89 !

PR #244, opened 2023-11-14
Utilize useRef to clear form data

hieungo89:
Clear form data after adding:

- Goals
- Subgoals/benchmarks


hieungo89:
Revised to PR #245.

PR #266, opened 2023-11-18
Add user to the staff list on the frontend instead of the backend

amantri:


codetheweb (approved):

hieungo89 (commented):
You will need to adjust some tests again since the previous getMyParas backend method also count self as a para. I believe the file is in `case_manager.test.ts`.

codetheweb (approved):

PR #268, opened 2023-11-21
Add breadcrumbs

hieungo89:
- Added a simple breadcrumbs layout using the address bar.
- Refactored some of the navbar logic grouping desktop view and mobile/tablet view.

Screenshot 2023-11-21 at 4 10 43 PM


src/components/breadcrumbs/BreadcrumbsNav.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const { data } = trpc.student.getStudentById.useQuery(
+      { student_id: paths[2] },
+      { enabled: Boolean(paths[2]), retry: false }
+    );
+    personData = data;
+  } else if (paths[1] === "staff" && paths[2]) {
+    const { data } = trpc.para.getParaById.useQuery(
+      { user_id: paths[2] },
+      { enabled: Boolean(paths[2]), retry: false }
+    );
```
codetheweb, line 27: this will probably break at some point, one of React's rules is that you can never change the hooks you call after the first render

so instead, you probably want to remove the `if` condition and put the same condition in the `enabled` option of the queries

hieungo89: Thanks for pointing it out @codetheweb! I fixed the logic with https://github.com/sfbrigade/compass/pull/268/commits/c779816aa46880c7a850b169b6d2315fa3535433 and https://github.com/sfbrigade/compass/pull/268/commits/30d07c426351639700082ddcfaaba1ce2a61b3b1. Please Review

src/components/breadcrumbs/BreadcrumbsNav.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const { data: para } = trpc.para.getParaById.useQuery(
+    { user_id: paths[2] },
+    { enabled: Boolean(paths[2] && paths[1] === "staff") }
+  );
+
+  const personData: Student | Para | undefined = student || para;
+
+  // An array of breadcrumbs fixed to students/staff as the first index. This will be modified depending on how the address bar will be displayed.
+  const breadcrumbs = paths.map((path, index) => {
+    console.log("path ", path, index);
```
codetheweb, line 30:
```suggestion
```

codetheweb (approved):

src/components/breadcrumbs/BreadcrumbsNav.tsx
hieungo89: https://github.com/sfbrigade/compass/pull/268/commits/20b75a74a8a05a59a78f07bcb2958051c3748562 - Removed

Issue #273, opened 2023-12-05
Create Design Systems folder for Reusable MUI components [ENG]


tessathornberry:
Buttons
Input
Modal
Dropdown
Breadcrumbs
 Search component
 Typography
 Progress bar
 Checkbox
 Tabs (style not finalized)

francisli:
This seems done in #275

PR #275, opened 2023-12-12
Added placeholder files and instruction documentation for Design System components

tessathornberry:
#273

Currently, the placeholders are generic (e.g. "Button") but that may add the slightest friction to coding in the future. For example, if someone imports an MUI "Button" as well as the Design Systems "Button" into a file and don't name the DS button something different on import, we may get an error. If we want things to be consistent or make it simple, I am happy to correct this with a prefix or suffix of "DS", like:

* ButtonDS
* DSButton

Please let me know what you think!


BrettEastman (approved):
The placeholders look fine, and it makes sense to create and organize a folder of design system components in this way. I like the titles as they are - no need to add "DS" as they are in the design_system folder.

PR #277, opened 2024-02-01
Update button module to override MUI styling and apply to all button instances 

hannahmcg:
In this PR, I essentially just took away the sx/individual styles within the MUI buttons and added more styling elements to the button css module so that any innate MUI styling elements would be overriden. Currently, the css module is able to make all the buttons look the same, but we still use both html and MUI buttons in our files. I didn't want to make any decisions yet on which one we could use as the standard, and would be happy to make a PR later to further standardize the button component!


codetheweb (approved):
awesome! I feel like this type of work is often thankless but much appreciated

BrettEastman (commented):
Yeah, great work, Hannah! Your work is definitely appreciated. Note, you can also cut the buttonSX const too: line 43 in staff/[user_id].tsx. It was a part of the sx of that previous button version.

PR #279, opened 2024-02-06
Update global color variables #263

BrettEastman:
### Updated all colors to css variables
- Checked all css variables in globals.css to make sure they represent Figma design system colors.
- Went through all instances of each hex code and changed to color variables, making educated guesses when coming across a hex color value not in the design system.
- Went through each instance of rgb colors and changed to color variables.
- Updated after [PR for ticket #277](https://github.com/sfbrigade/compass/pull/277) approved and merged.


src/components/subgoal/Subgoal.tsx
```diff
@@ -43,7 +43,7 @@ const Subgoals = ({ subgoal }: SubgoalProps) => {
                 flex: "flex-end",
                 width: "auto",
                 padding: "20px 10px",
-                backgroundColor: "#fff ",
+                backgroundColor: "var(--grey-100)",
                 borderWidth: "1px",
                 borderColor: "#20159E",
```
hieungo89: This border color can also be converted. Everything else looks good.

src/styles/Button.module.css
tessathornberry, line 1: Is there a reason we have this secondary file in styles instead of using src/components/design_system/button/Button.module.css?

tessathornberry (dismissed):
For the design system components/shared designs, we have the folders for reusable components in src/components/design_system/. These SHOULD be sufficient to store code for the actual reusable design components themselves or only the CSS to apply to html and MUI components with className. 
In what situation might we need two of these files?

src/styles/Button.module.css
BrettEastman: Thank you for spotting this! That was a duplicate left over in my branch from the previous system we had prior to February. I have now removed it.

tessathornberry (approved):
Great job!

PR #283, opened 2024-02-12
Add typography and paper elevations to theme

codetheweb:
Added typography & paper variants to theme to centralize styling. Also added Storybook pages to preview both.


hannahmcg (approved):
Looks great!! Nice work with storybook too

PR #294, opened 2024-02-23
Add scalable font size to global and adjust font size of other files.

hieungo89:
#293

Font size at .875rem [14px] at screens lower than 480px.
Font size at .9375rem [15px] at screens above 480px.
Font size at 1rem [16px] at screens above 768px.
Add reusable font-size classes including small and x-small sizes.

Following [Typography chart ](https://www.figma.com/file/m09znscRXNqSziAiEbNLTf/Compass-Designs?type=design&node-id=12388-48833&mode=design&t=ghaGyca0JE0N57m0-4)
Typography

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -138,7 +138,7 @@ p {
 /* Small Devices, Phones & Tablet [long screen] */
 @media only screen and (min-width: 480px) {
   :root {
-    --base-size: .925rem;
+    --base-size: .9375rem;
   }
 }
 
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -49,7 +49,7 @@
   --grey-100: #ffffff;
 
   /* Easy Text Size Scaling */
-  --base-size: .875rem; /* equivalent to 14px */
+  --base-size: 0.875rem; /* 14px */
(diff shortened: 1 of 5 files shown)


src/styles/globals.css
```diff
@@ -78,17 +97,64 @@ textarea {
...
+  font-size: var(--sm);
+}
+
+.xs {
+  font-size: var(--xs);
+}
+
+p {
+  font-size: var(--base-size);
+}
```
codetheweb, line 141: this is ok as a stopgap since we've already used `<h*/>` elements everywhere but might be a bit better to define these sizes in MUI's theme (I think they might already be there) and use `<Typography/>` elements

codetheweb (approved):

src/styles/globals.css
hieungo89: I'm not familiar with the MUI theme and how defining it can override these elements. In any case, these can be used as fallback if the MUI theme is not defined right?

codetheweb: yep, just trying to avoid duplicated definitions

codetheweb (approved):

PR #299, opened 2024-03-05
Minor CSS Fixes (sticky nav and layout)

hannahmcg:
Made navbar sticky and layout background now extends past screen height when content overflows


codetheweb (approved):

hieungo89 (changes requested):
We are using a mobile-first design. Instead of changing it to max-width, you can keep the min-width and change line 13 margin-left: 0;, then at line 43 min-width:992, the margin-left: 200px; kicks in.

hieungo89 (approved):

Issue #212, opened 2023-10-16
Make IEP date editable [ENG]


tessathornberry:
- [x] Add start date and end date to edit student inputs
- [x] Add start date and end date to tRPC mutations
- [x] Add start date and end date to API/routes for updating the database
- [x] Add ava test for back end API in routes

### NOTE: check for date agreement between user inputs and resulting saved edited dates - date-fns formatting will likely be required to divert dates from UTC formatting

tessathornberry:
Figma:
https://www.figma.com/file/m09znscRXNqSziAiEbNLTf/Compass-Designs?type=design&node-id=9351%3A29474&mode=design&t=cM4m4Vu4Zzq4iVFk-1

PLEASE NOTE: this design is under revision and the updated version will be provided soon.
Designer: Neha - reach out with any questions

tessathornberry:
Will work on this with Katrina once Modal is Approved by designers

tessathornberry:
This ticket only makes editing the IEP date possible. The remodeling of the EDIT page to a modal is in ticket #302

PR #305, opened 2024-03-29
Added editIEP to the front and back end and wrote test. Future builds will include these in a modal.

tessathornberry:
This closes ticket #212 

* Iep layout is closer to design systems layout
* Button for archive is moved out to Student instead of inside edit.
* Edit items have been put in order of creation
* Edit IEP options entered into form (CSS will be finalized/updated with Modal creating in next ticket)
* Edit IEP trpc function created
* Ava test for editIep created and passed

editiep

### DIFF
--- a/src/backend/routers/student.test.ts
+++ b/src/backend/routers/student.test.ts
@@ -81,49 +81,55 @@ test("addIep and getIep", async (t) => {
   t.deepEqual(got[0].end_date, added.end_date);
 });
 
-// ! this is the edit test
 test("editIep", async (t) => {
   const { trpc, seed } = await getTestServer(t, {
     authenticateAs: "case_manager",
   });
 
-  const start_date = new Date("2023-01-01");
-  const end_date = new Date("2023-12-31");
+  // * must add student this way to populate the assigned_case_manager_id
+  await trpc.case_manager.addStudent.mutate({
+    first_name: seed.student.first_name,
+    last_name: seed.student.last_name,
+    email: seed.student.email,
+    grade: seed.student.grade,
+  });
 
-  const added = await trpc.student.addIep.mutate({
-    student_id: seed.student.student_id,
-    start_date: start_date,
-    end_date: end_date,
+  const myStudents = await trpc.case_manager.getMyStudents.query();
+  const student_id = myStudents[0].student_id;
+
+  const iep = await trpc.student.addIep.mutate({
+    student_id: student_id,
+    start_date: new Date("2023-01-01"),
+    end_date: new Date("2023-12-31"),
   });
 
   let got = await trpc.student.getIeps.query({
-    student_id: seed.student.student_id,
+    student_id: student_id,
   });
 
   t.is(got.length, 1);
-  t.is(got[0].student_id, seed.student.student_id);
-  t.deepEqual(got[0].start_date, added.start_date);
-  t.deepEqual(got[0].end_date, added.end_date);
+  t.is(got[0].student_id, iep.student_id);
+  t.deepEqual(got[0].start_date, iep.start_date);
+  t.deepEqual(got[0].end_date, iep.end_date);
 
   const updated_start_date = new Date("2023-03-02");
   const updated_end_date = new Date("2024-03-01");
 
   await trpc.student.editIep.mutate({
-    student_id: seed.student.student_id,
+    student_id: student_id,
     start_date: updated_start_date,
     end_date: updated_end_date,
   });
 
   got = await trpc.student.getIeps.query({
-    student_id: seed.student.student_id,
+    student_id: student_id,
   });
 
   t.is(got.length, 1);
-  t.is(got[0].student_id, seed.student.student_id);
-  t.notDeepEqual(got[0].start_date, start_date);
-  t.notDeepEqual(got[0].end_date, end_date);
+  t.is(got[0].student_id, student_id);
+  t.notDeepEqual(got[0].start_date, iep.start_date);
+  t.notDeepEqual(got[0].end_date, iep.end_date);
 });
-// ! above is the edit test
 
 test("getActiveStudentIep - return only one iep object", async (t) => {
   const { trpc, seed } = await getTestServer(t, {
--- a/src/backend/routers/student.test.ts
+++ b/src/backend/routers/student.test.ts
@@ -81,6 +81,39 @@ test("addIep and getIep", async (t) => {
   t.deepEqual(got[0].end_date, added.end_date);
 });
 
+test("editIep", async (t) => {
+  const { trpc, seed } = await getTestServer(t, {
+    authenticateAs: "case_manager",
+  });
+
+  const start_date = new Date("2023-01-01");
+  const end_date = new Date("2023-12-31");
+
+  const added = await trpc.student.addIep.mutate({
+    student_id: seed.student.student_id,
+    start_date: start_date,
+    end_date: end_date,
+  });
+
+  const updated_start_date = new Date("2023-03-02");
+  const updated_end_date = new Date("2024-03-01");
+
+  const updated = await trpc.student.editIep.mutate({
+    student_id: seed.student.student_id,
+    start_date: updated_start_date,
+    end_date: updated_end_date,
+  });
+
+  const got = await trpc.student.getIeps.query({
+    student_id: seed.student.student_id,
+  });
+
+  t.is(got.length, 1);
+  t.is(got[0].student_id, seed.student.student_id);
+  t.deepEqual(got[0].start_date, updated.start_date);
+  t.deepEqual(got[0].end_date, updated.end_date);
+});
+
 test("getActiveStudentIep - return only one iep object", async (t) => {
   const { trpc, seed } = await getTestServer(t, {
     authenticateAs: "case_manager",
--- a/src/backend/routers/student.test.ts
+++ b/src/backend/routers/student.test.ts
@@ -81,6 +81,7 @@ test("addIep and getIep", async (t) => {
   t.deepEqual(got[0].end_date, added.end_date);
 });
 
+// ! this is the edit test
 test("editIep", async (t) => {
   const { trpc, seed } = await getTestServer(t, {
     authenticateAs: "case_manager",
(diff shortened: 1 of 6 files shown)


src/components/design_system/button/Button.module.css
```diff
@@ -57,6 +57,27 @@
...
 
+.tertiary {
+  display: block;
+  color: var(--primary-50);
+  padding: 10px 12px;
+  border-radius: 8px;
+  cursor: pointer;
+  text-decoration: underline;
+  text-align: center;
+  font-size: 16px;
```
hieungo89, line 68: I would advise having font size as either `1 rem` or the global size of `var(--h6)` so that it's scalable with smaller screens.

src/backend/routers/student.ts
```diff
@@ -64,6 +64,42 @@ export const student = router({
...
+        start_date: z.date(),
+        end_date: z.date(),
+      })
+    )
+    .mutation(async (req) => {
+      const { student_id, start_date, end_date } = req.input;
+      const { userId } = req.ctx.auth;
+
+      // Check if the student exists and if the case manager is assigned to the student
+      const existingStudent = req.ctx.db
```
hieungo89, line 83: I think you might need to also add `.execute()` to run this method.

src/backend/routers/student.ts
```diff
@@ -64,6 +64,42 @@ export const student = router({
...
+      const result = await req.ctx.db
+        .updateTable("iep")
+        .set({
+          start_date,
+          end_date,
+        })
+        .where("student_id", "=", student_id)
+        .returningAll()
+        .executeTakeFirstOrThrow();
+      return result;
```
hieungo89, line 100: Do we need to retrieve data directly back from this `editIep` method? If not, then a return statement is not needed or a result variable. It seems like the frontend method is already retrieving the new data after the edition is saved.

src/backend/routers/student.test.ts
```diff
@@ -81,6 +81,39 @@ test("addIep and getIep", async (t) => {
...
 
+test("editIep", async (t) => {
+  const { trpc, seed } = await getTestServer(t, {
+    authenticateAs: "case_manager",
+  });
+
+  const start_date = new Date("2023-01-01");
+  const end_date = new Date("2023-12-31");
+
+  const added = await trpc.student.addIep.mutate({
```
hieungo89, line 92: Lint advised removing the const since `added` is not being used elsewhere.

src/backend/routers/student.test.ts
```diff
@@ -81,6 +81,39 @@ test("addIep and getIep", async (t) => {
...
+    end_date: updated_end_date,
+  });
+
+  const got = await trpc.student.getIeps.query({
+    student_id: seed.student.student_id,
+  });
+
+  t.is(got.length, 1);
+  t.is(got[0].student_id, seed.student.student_id);
+  t.deepEqual(got[0].start_date, updated.start_date);
```
hieungo89, line 113: I would suggest doing a check for the start & end dates before the updated dates, and then showing the updated start & end dates are not the same as the previously added dates. You also don't need a const for `updated` since it shouldn't return anything. You should get your data with the `.getIeps.query()` instead.

src/backend/routers/student.ts
```diff
@@ -64,6 +64,42 @@ export const student = router({
...
+      const { student_id, start_date, end_date } = req.input;
+      const { userId } = req.ctx.auth;
+
+      // Check if the student exists and if the case manager is assigned to the student
+      const existingStudent = req.ctx.db
+        .selectFrom("student")
+        .selectAll()
+        .where("student_id", "=", student_id)
+        .where("assigned_case_manager_id", "=", userId);
+      if (!existingStudent) {
```
KCCPMG, line 88: I could be wrong and I haven't checked this, but if you're using selectAll, then doesn't this query return an array? If so, then I believe that existingStudent will always be truthy as an array with length of 0. I think you want to check if existingStudent.length === 0 rather than the current check, but either way I'd recommend adding tests to make sure that the error is being thrown for a missing or invalid student.

src/backend/routers/student.test.ts
KCCPMG: Just came here to say the same thing. I was playing around with this and tried to compare updated.start_date with updated_start_date and got the following difference showing in the test log:

  Difference (- actual, + expected):

  - Date 2023-03-01 08:00:00 UTC {}
  + Date 2023-03-02 00:00:00 UTC {}

src/backend/routers/student.ts
hieungo89: `selectAll()` would be selecting all the key:value pairs from all the matches in the db. However, I mentioned above that there might not be any returned data due to not having the `execute()` statement. I do agree with you that checking the length is more accurate.

src/components/design_system/button/Button.module.css
tessathornberry: Great call, @hieungo89 ! I reused the code for the secondary button and don't know why that became pixels. The secondary button font size is var(--base-size). Would that work?

src/backend/routers/student.ts
tessathornberry: Good call!

src/backend/routers/student.ts
tessathornberry: Great call, it works well without it and I agree, thank you!

src/backend/routers/student.test.ts
tessathornberry: @hieungo89 , "updated" inserts the data into the iep keys, it doesn't retrieve anything. I can make "got" a let variable and reuse it after the data has been edited. re: your previous comment, if I make a test for checking the starting values, I will use "added" for that test. I can take out "updated" and simply have the await portion of the mutation of editIep, re-calling the "getIeps" with the "got" variable so I can use the new value in "notDeepEqual" tests.
Let me know if what I do is close to what you had in mind? I'll push it up once done with all comments.

src/backend/routers/student.ts
tessathornberry: Good call, @KCCPMG and @hieungo89 . I snagged much of this from the addIep function, which has the same issues.

tessathornberry: FYI, 

how about 

```
if (!existingStudent[0].student_id) {
        throw new Error("Student not found under this Case Manager");
      }
```

since existingStudent =

```
[
  {
    student_id: '9eaa9500-a2a9-4704-b917-7b8c2f28f104',
    first_name: 'student',
    last_name: 'one',
    email: 'one@student.com',
    assigned_case_manager_id: 'a5663b61-691b-49d6-8322-1b3e3f4ba79f',
    grade: 5
  }
]
```

Length checks would be dicey on undefined properties since it may cause a different kind of error.

tessathornberry: I noted that the major issue was how the seed data was populating the student table, ensuring the assigned_case_manager_id was NOT null.

tessathornberry: > `selectAll()` would be selecting all the key:value pairs from all the matches in the db. However, I mentioned above that there might not be any returned data due to not having the `execute()` statement. I do agree with you that checking the length is more accurate.

[RE: Kysley](https://kysely.dev/docs/category/select):
```
const persons = await db
  .selectFrom('person')
  .selectAll()
  .execute()
```
translates to:
```
SELECT * FROM "person";
```
You have to have a selectFrom and a select WHAT, even if you're not returning anything. Also, the `execute()` can be any of the execute commands and it will return something, even if you do not need anything returned.

I ended up doing the conditional on `!existingStudent[0]`: please let me know if this works for you.

hieungo89: I checked, and the return from `existingStudent` is either array with 0-1 data point. I believe this should be fine. 👍

src/pages/students/[student_id].tsx
```diff
@@ -61,20 +56,32 @@ const ViewStudentPage = () => {
...
+
     editMutation.mutate({
       student_id: student.student_id,
       first_name: data.get("firstName") as string,
       last_name: data.get("lastName") as string,
       email: data.get("email") as string,
       grade: Number(data.get("grade")) || 0,
     });
+
+    editIepMutation.mutate({
```
Comment (hieungo89) [inline src/pages/students/[student_id].tsx:79]: I noticed that there is no checker to ensure that the end date is always after the start date. Do you want to add that logic in this PR or make it a separate issue?

Comment (hieungo89) [reply src/pages/students/[student_id].tsx:79]: On the input for `endDate` line 303, if you put a `min={startDate}`, then you cannot select an end date before the start date. Also, lines 290 and 306 have reference to the states, but the states did not change, therefore those values remain empty strings. If you add this to your handleEditState, your values will populate the dates for the date picker.

`  const handleEditState = () => {
    setViewState(VIEW_STATES.EDIT);
    if (activeIep) {
      setStartDate(activeIep.start_date.toISOString().slice(0, 10));
      setEndDate((activeIep.end_date).toISOString().slice(0, 10));
    }
  };`

Comment (hieungo89) [reply src/pages/students/[student_id].tsx:79]: Screenshot 2024-04-04 at 9 46 23 PM

hieungo89:
Test looks good!

src/backend/routers/student.test.ts
```diff
@@ -81,6 +81,56 @@ test("addIep and getIep", async (t) => {
...
+    end_date: updated_end_date,
+  });
+
+  got = await trpc.student.getIeps.query({
+    student_id: student_id,
+  });
+
+  t.is(got.length, 1);
+  t.is(got[0].student_id, student_id);
+  t.notDeepEqual(got[0].start_date, iep.start_date);
```
KCCPMG, line 130: I don't think this part of the test is really capturing what you want since it's just checking to see if the date is changed, especially if there's still weird date stuff going on.

src/backend/routers/student.test.ts
```diff
@@ -81,6 +81,56 @@ test("addIep and getIep", async (t) => {
...
+  t.notDeepEqual(got[0].end_date, iep.end_date);
+});
+
```
KCCPMG, line 133: I think that it would be ideal to add a few other tests here to make sure that errors are being thrown in the right circumstances, specifically when an edit attempt is made on a non-existent student, and when an edit attempt is made on a student that is not assigned to the case manager. This would be beyond the standard of what I believe has been done in similar situations though, so I wouldn't consider it to be essential by any stretch.

KCCPMG (changes requested):

src/backend/routers/student.test.ts
tessathornberry: I agree, in part, and have added an isDeepEqual back in as well as wrapping the new Date strings with a parseISO from date-fns, which adds more specific time-zone-related information to the date string and handles the date mismatch in the test environment. (It is already correct in the frontend handling)

As for the tests re: the existing student, that is already handled in the editIep function itself and as you say, is outside of the scope of this test.

src/pages/students/[student_id].tsx
Comment (tessathornberry) [reply src/pages/students/[student_id].tsx:84]: Thanks so much for that, @hieungo89 ! I added those initial values in the handleEdit and the min in the endDate, as well as an onChange in the startDate that keeps the endDate min dynamic.

tessathornberry:
editIep

src/backend/routers/student.test.ts
KCCPMG: Right, the thing that I'm suggesting is adding an additional test to make sure that the error is being thrown, rather than expanding the editIep test. If you wanted to do that, here's one that I believe works:

`test("editIep correctly throws for non-existent student", async (t) => {
  const { trpc, seed } = await getTestServer(t, {
    authenticateAs: "case_manager",
  });

  const updated_start_date = new Date(parseISO("2023-03-02"));
  const updated_end_date = new Date(parseISO("2024-03-01"));

  await t.throwsAsync(
    trpc.student.editIep.mutate({
      student_id: randomUUID(), //random uuid not matching any valid document
      start_date: updated_start_date, 
      end_date: updated_end_date
    }), {message: "Student not found under this Case Manager"}
  );
})`
Like I said above, not essential, so I'll drop this here and mark it as resolved.

Also the date fix looks good!

KCCPMG (approved):

hieungo89 (approved):

PR #309, opened 2024-04-08
Fix add staff

KCCPMG:
Replaces two subsequent trpc calls with one, with the logic for that call broken out into several functions in a new helper file. Calls for the creation of the invitation email but does not await the result of the promise, so email failures will no longer hold up a response from being sent to the client. Adding staff should be clear and straightforward for a user, but any Postgres error that causes the mutation to fail (duplicate email) could still benefit from a toast notification as to the cause of the error.


codetheweb (approved):

katconnors:
Functionality: Button/add functionality works
Testing: All tests passed on my end
Comments in the code were helpful, so keep up the good work on that.
There are a few github warnings from the lint check (mainly warnings for items that aren't being used), so you could look into those for code cleanup contenders.
I don't have suggestions for other changes at the moment.

PR #310, opened 2024-04-10
Update .env.example

tessathornberry:
Added note regarding solution for postgres docker connection issues


katconnors:
Looks good! Thanks for adding this.

katconnors (approved):

Issue #313, opened 2024-04-12
Edit Error Text to Instruct User to enter in proper email format [ENG]


katconnors:
As is, when a user enters the incorrect email format, the following error message appears. However, it is misleading.  Currently, the logic will lead to displaying the below error in all error instances.
![image](https://github.com/sfbrigade/compass/assets/32425204/c7c4bc3a-14d1-4e2a-8828-7890389eff01)

- [x] Update error logic in compass>src>pages>students>index.tsx to have a different error text for the above case.

PR #320, opened 2024-05-03
Resolution to Issue #313: Error alert for invalid email entry

katconnors:
Issue313

There is now an instructive alert that will appear when the wrong email format is submitted.
(see issue #313 for the previous state)
Consulted trpc docs, but didn't find a more straightforward way to access the message for the error- that is a potential area for improvement.

### DIFF
--- a/src/pages/students/index.tsx
+++ b/src/pages/students/index.tsx
@@ -16,19 +16,19 @@ const Students = () => {
     onError: (errortraits) => {
       // errortraits allows one to access validation, code, message, and path
       // JSON.parse is utilized because errortraits.message is a string
-      const formattederrortraits = JSON.parse(errortraits.message) as {
-        validation: string;
-        code: string;
-        message: string;
-        path: string[];
-      }[];
+      try {
+        const formattederrortraits = JSON.parse(errortraits.message) as {
+          validation: string;
+          code: string;
+          message: string;
+          path: string[];
+        }[];
 
-      if (formattederrortraits[0].message == "Invalid email") {
-        alert("The provided email is in the incorrect format- please edit.");
-      }
-
-      //  can later insert other error messages here, as needed
-      else {
+        if (formattederrortraits[0].message == "Invalid email") {
+          alert("The provided email is in the incorrect format- please edit.");
+        }
+      } catch {
+        //  can later insert other error messages here, as needed
         alert(
           `This student is already assigned to a case manager. Please check your roster if the student is already there. Otherwise, this student is with another case manager.`
         );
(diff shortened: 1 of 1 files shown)


KCCPMG (changes requested):
This appears to work fine for the email format error, but does not correctly handle the error from a student already being assigned. It looks to me like the reason for this is that an incorrect email format will trigger a Zod error, where you have the error message as correctly being parsed as an array from the JSON, but the error message coming from other errors in the Kysely query will just be a string. This can be solved on its own by using a try/catch block and putting the JSON.parse call in the try and putting the other alert in the catch block. I would also recommend a bit of a refactor on the router side to catch and throw errors that we can control (the current actual error being thrown on a conflict is simply that the query returns no results), but that's really outside the scope of this issue and could be a simple issue on its own.

src/pages/students/index.tsx
```diff
@@ -16,19 +16,19 @@ const Students = () => {
...
       // errortraits allows one to access validation, code, message, and path
       // JSON.parse is utilized because errortraits.message is a string
-      const formattederrortraits = JSON.parse(errortraits.message) as {
-        validation: string;
-        code: string;
-        message: string;
-        path: string[];
-      }[];
+      try {
+        const formattederrortraits = JSON.parse(errortraits.message) as {
```
KCCPMG, line 20: Would recommend renaming "errortraits" to "e", "err", or "error" to fit more with standard convention, would similarly recommend renaming "formattederrortraits" based on that and using camelCase for the new name or this one if you keep it.

src/pages/students/index.tsx
```diff
@@ -16,19 +16,19 @@ const Students = () => {
...
 
-      if (formattederrortraits[0].message == "Invalid email") {
-        alert("The provided email is in the incorrect format- please edit.");
-      }
-
-      //  can later insert other error messages here, as needed
-      else {
+        if (formattederrortraits[0].message == "Invalid email") {
+          alert("The provided email is in the incorrect format- please edit.");
+        }
```
KCCPMG, line 29: I would recommend adding an else statement here for if there's an error that matches the format you're checking against, but the error doesn't match the message you're checking against. Basically just something like 

else {
    alert("Something went wrong");
}

This is another thing that I think is probably good practice rather than absolutely essential given that the current try/catch code that you have is capturing all of the error contingencies we're currently aware of.

KCCPMG (approved):
I have a couple more quick suggestions, but this looks to me like it works and I'll approve whether you want to take those into account or not.

Issue #316, opened 2024-04-15
Min IEP Date Check/Fix [ENG]


katconnors:
Min as an Input Prop may not be working as intended (pages/students/[student_id]). 

- [x] Check behavior of pages/students/[student_id] to see if you can change the end IEP date to be less than the start IEP date
- [x] If yes, edit/add a check to have the end IEP date be, at a minimum, the IEP start date

katconnors:
It seems as though the faulty behavior exists after the IEP is created/in edit mode,

katconnors:
Will work on this in edit-iep-check branch

PR #335, opened 2024-05-30
Fixed check for iep end date- resolution to [ENG] #316 issue

katconnors:
Change to allow for proper check of the min IEP date.
Now, the end IEP dates are greyed out to prevent selection of a date that is less than the start IEP date.

Before:
image

After:
image


KCCPMG (approved):
Looks like this does the trick, might be worth adding a check inside the schema as well, but that can be its own (maybe a good first) issue

PR #336, opened 2024-06-03
Create new ProgressBar component

BrettEastman:
### Changes:
- Created new **ProgressBar** component in components/design_system/progressBar according to specifications in the official design system.
- Deleted old ProgressBar component.
- Added color palette to MUI theme, with the basic colors from the design system (`primary.main` and `primary.light` were used in the creation of this component).

### Demonstration:
To view, go to the Settings page. Under the "Under Construction" notice, I put the new Progress Bar component, set to 50%. Note: Progress Bar has not yet been implemented to any other components or files in the codebase, hence the demonstration. I will remove this from the Settings page once the PR is approved.
Screenshot 2024-06-03 at 3 53 27 PM


src/theme.ts
```diff
@@ -3,6 +3,26 @@ import { createTheme } from "@mui/material";
 const { breakpoints } = createTheme();
 
 export const theme = createTheme({
+  palette: {
+    primary: {
+      main: "#3023b8",
+      light: "#9b93f1",
+      dark: "#080155",
```
katconnors, line 10: Is there currently a way to switch the theme on the site?

katconnors (approved):
Looks good to me- just had a small question for my own understanding (see other comment).

src/theme.ts
BrettEastman: No, there is no light/dark theme option currently on the site. Those are just three selections from the design system primary colors. Really, I just needed primary.main for the ProgressBar component, but it might be handy to have those in the future. I modeled it after this: [mui](https://v3.mui.com/customization/themes/)

Issue #324, opened 2024-05-09
Create Updated create benchmark flow [ENG]


nikobooth:
Neha has updated the create benchmark flow to collect all the information needed for a staff member to understand the data collection task. We have a ticket in progress to deploy the previous design. Need to decide how to get from what's in progress to the new design.

Yun, Katrina, Connor to discuss.

New design [Figma](https://www.figma.com/file/m09znscRXNqSziAiEbNLTf/Compass-Designs?type=design&node-id=13077-24231&mode=design&t=ev6EhVxqxp2sQTjX-11)

Following checklist is not exhaustive:
- [ ] Breakdown Into separate tickets (Connor is in progress on this writeup, tbd discussion)
- [ ] https://github.com/sfbrigade/compass/issues/325

nikobooth:
Main updates in new design:

- Tabs in goal view for `All` & `Completed` benchmaks
- New inputs: Attemps per trial & Number of trials
- Creation is now a two step flow with a new stepper component

KCCPMG:
![Image](https://github.com/sfbrigade/compass/assets/39074889/255b1591-9571-4f21-a980-051b46c080b0)


![Image](https://github.com/sfbrigade/compass/assets/39074889/ef6e411f-d252-41d7-90b9-b5f1c8f4e514)


![Image](https://github.com/sfbrigade/compass/assets/39074889/95bbcc65-a90c-455b-9b67-0fbbf409e588)


![Image](https://github.com/sfbrigade/compass/assets/39074889/7db9b6f2-48a5-4f11-b827-5a6140bfbf4a)


![Image](https://github.com/sfbrigade/compass/assets/39074889/ec2c0c3e-f6a2-48d5-a8de-06ba62283a6d)

chengtchris1:
Working on creating the Benchmark list tile. Here is the branch I created
https://github.com/sfbrigade/compass/tree/Created-benchmark-view-new-design

KCCPMG:
Created issue #339 to handle the database update aspect. This needs to be handled before the styling can be done, as the data being passed to pages and components will be different from the current version.

KCCPMG:
Issue #339 closed

KCCPMG:
Issue [#341](https://github.com/sfbrigade/compass/issues/341) created to address `students/[student_id]/goals/[goal_id]` page, assigned to @chengtchris1 .

KCCPMG:
[...]

PR #340, opened 2024-06-06
Fix benchmark schema update

KCCPMG:
Updates initial_migrations, routers and tests for routers/iep and routers/para, and several downstream components. When this is approved and merged, remaining work on #324 can be divvied up and assigned.


src/backend/routers/iep.test.ts
```diff
@@ -23,24 +23,30 @@ test("basic flow - add/get goals, subgoals, tasks", async (t) => {
...
+    status: "In Progress",
     description: "subgoal 1",
+    setup: "Setup here",
     instructions: "instructions here",
-    target_max_attempts: 5,
     materials: "materials",
-    target_level: 100,
-    baseline_level: 20,
+    target_level: 85,
+    baseline_level: 60,
```
katconnors, line 32: May be good to add a test for current_level as well?

katconnors: That makes sense! Thanks. I think a comment (or no changes) should be okay.

katconnors (approved):
New fields look good on the site, and the code looks good to me.

src/backend/routers/iep.test.ts
KCCPMG: Perfect! I added a comment there and to the initial_migrations to clarify that. I kept the test I added this morning and made sure to check that the value of the returned current_level was null, to make sure it wasn't getting a value but was still being included in the returned object.

Issue #334, opened 2024-05-30
Lint check errors in [student_id].tsx


katconnors:
- [ ] Resolve errors in compass/src/pages/students/[student_id].tsx: "Unsafe assignment of an `any` value"

katconnors:
PR here https://github.com/sfbrigade/compass/pull/351

PR #351, opened 2024-07-16
Resolution of any value error within [student_id] - Issue #334

katconnors:
See issue here: https://github.com/sfbrigade/compass/issues/334
Import of the custom/project Modal component was not occurring.
Note that the staff page does not utilize the same Modal, so the reference doesn't need updates.
Co-authored-by: Yun Whitman <savannahwhitman1@gmail.com>

### DIFF
--- a/src/pages/students/[student_id].tsx
+++ b/src/pages/students/[student_id].tsx
@@ -10,11 +10,13 @@ import $Image from "../../styles/Image.module.css";
 import $button from "@/components/design_system/button/Button.module.css";
 import $Form from "../../styles/Form.module.css";
 import $input from "@/styles/Input.module.css";
-import $Modal from "@/components/design_system/modal/Modal.module.css";
+import $CompassModal from "../../components/design_system/modal/CompassModal.module.css";
 import $StudentPage from "../../styles/StudentPage.module.css";
 
 import * as React from "react";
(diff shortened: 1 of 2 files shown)


src/pages/students/[student_id].tsx
```diff
@@ -10,12 +10,14 @@ import $Image from "../../styles/Image.module.css";
...
 import $Form from "../../styles/Form.module.css";
 import $input from "@/styles/Input.module.css";
-import $Modal from "@/components/design_system/modal/Modal.module.css";
+import $CompassModal from "../../components/design_system/modal/CompassModal.module.css";
 import $StudentPage from "../../styles/StudentPage.module.css";
 
 import * as React from "react";
 import Typography from "@mui/material/Typography";
 
+// test comment
```
Comment (KCCPMG) [inline src/pages/students/[student_id].tsx:19]: Probably don't need this anymore?

KCCPMG (approved):
Looks good. I linked this PR to issue 334, but when I was looking for it, I didn't actually see the original error that #334 was referencing- maybe it was cleaned up in an earlier revision? At any rate I think this is good to go assuming there are no merge issues.

src/pages/students/[student_id].tsx
Comment (katconnors) [reply src/pages/students/[student_id].tsx]: Thanks Connor! deleted this extraneous comment

Issue #342, opened 2024-06-07
Create Updated Benchmark Flow: Update [goal_id]/create page for benchmark creation [ENG]


KCCPMG:
Update `students/[student_id]/goals/[goal_id]/create` page to accommodate the two-step benchmark creation form as per the new designs. The Figma designs can be found [here](https://www.figma.com/file/m09znscRXNqSziAiEbNLTf/Compass-Designs?type=design&node-id=13077-24231&mode=design&t=ev6EhVxqxp2sQTjX-11), and the images can be seen below:

![Image](https://github.com/sfbrigade/compass/assets/39074889/ef6e411f-d252-41d7-90b9-b5f1c8f4e514)
![Image](https://github.com/sfbrigade/compass/assets/39074889/95bbcc65-a90c-455b-9b67-0fbbf409e588)

Summary of Potential Action Items

- [ ] Implement in the Design Systems a "Stepper" component. [MaterialUI provides this](https://mui.com/material-ui/react-stepper/#customized-horizontal-stepper), though I'm not sure if one of these formats will match the design. If necessary, this can be extracted to a separate issue and replaced here with a placeholder.
- [ ] Redo the submission form so that it matches the design, including that it appears across two "pages" within the page. This logic should likely be extracted to its own component, and all of this will probably be a large task.

nickvisut:
Stepper component ticket here: https://github.com/sfbrigade/compass/issues/325

PR #352, opened 2024-07-17
Update benchmark flow: Update [goal_id]/create page for benchmark creation Ticket #342

BrettEastman:
### Redesign and Refactor Goal Creation Flow

**Description**:

This pull request implements a comprehensive redesign and refactoring of the goal creation flow, specifically focusing on the `[goal_id]/create.tsx` component, improving user experience, aligned with official designs.

Key Changes:

Restructured layout:
   - Implemented row-based design with title, subtitle, and required TextField components
   - Added actual titles and placeholder text from official designs
   - Converted TextField data to an array of objects, which is mapped over to render TextFields with appropriate titles and descriptions
   - Implemented logic for navigation buttons (next, cancel, back, create)
   - Added Frequency field to the page
   - Replaced progress bar with Stepper component
   - Implemented form validation logic to have an alert pop up if not all fields have been filled out

Implemented view state management:
- Added functionality to switch between first and second pages

TO DO in future PRs:
- Add frequency field to the schema


src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -57,102 +153,115 @@ const CreateBenchmarkPage = () => {
...
-              name="setup"
-            />
-
-            <Stack spacing={2}>
-              <Typography variant="h4">Metrics</Typography>
-              <Stack direction="row" spacing={4}>
-                <Stack direction="row" alignItems="center">
+        <Stack direction="row" spacing={4}>
+          {viewState === VIEW_STATES.BENCHMARK_PG_1 && (
+            <Stack spacing={4} px={3} pb={3} width={"100%"}>
```
Comment (KCCPMG) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:180]: This is a really cool approach, but it looks to me like when advancing to the second page, the content from VIEW_STATES.BENCHMARK_PG_1 is dropped, and then the data that filled out those text fields is lost. The thing I immediately think of to suggest is to render both "pages", but set visibility conditionally based on which page is selected. I don't know if that would cause an issue with FormData's ability to retrieve content, though. The other idea I would have would be to create a piece of state for each "page" that is set by collecting formData when clicking through to "Next" or to "Create Benchmark", then submitting the "page 1 state" and "page 2 state" together at the time of form submission. Just spitballing here, so feel free to do whatever works.

KCCPMG (changes requested):
Overall this is a great job of implementing the design, but we have to take care of some of the logic stuff here. Let me know whatever I can do to help.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
Comment (BrettEastman) [reply src/pages/students/[student_id]/goals/[goal_id]/create.tsx:207]: After our conversation, it ended up that, since the form is split into two view states, plan B was the best option - using the state object data for the subgoalMutation, rather than the FormData API.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -31,20 +61,86 @@ const CreateBenchmarkPage = () => {
...
+      description:
+        "Describe how staff should conduct a trial for this benchmark.",
+      label: "Instructions",
+      name: "instructions",
+      placeholder:
+        "What level of prompting is permitted, what specific actions staff should take, what they should be observing...",
+    },
+  ];
+
+  const textFieldData2 = [
```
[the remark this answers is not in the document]
Comment (BrettEastman) [reply src/pages/students/[student_id]/goals/[goal_id]/create.tsx]: I added back in the frequency field, and I left a TO DO comment above the subgoalMutation that we need to update the schema to accommodate this. So, it is collecting that data in the state object, but it isn't currently being sent anywhere.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -57,102 +180,160 @@ const CreateBenchmarkPage = () => {
...
+        {/* TO DO: Implement logic to check and see if required fields are filled out before allowing user to proceed to the next page */}
+        {viewState === VIEW_STATES.BENCHMARK_PG_2 && (
+          <Stack direction="row" spacing={2} p={4} justifyContent="end">
+            <button
+              onClick={() => setViewState(VIEW_STATES.BENCHMARK_PG_1)}
+              className={$button.secondary}
+            >
+              Back
+            </button>
+            <button type="submit" className={$button.default}>
```
Comment (KCCPMG) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:332]: Suggestion here to go with line 59:

`<button type="submit" className={$button.default} disabled={!formComplete}>`

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -13,23 +36,45 @@ const CreateBenchmarkPage = () => {
...
+      instructions: undefined,
+      baseline_level: undefined,
+      target_level: undefined,
+      attempts_per_trial: undefined,
+      number_of_trials: undefined,
+    });
+
+  const steps = ["Instructional Guidelines", "Data Collection Guidelines"];
+
   const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
```
Comment (KCCPMG) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:60]: Hopefully we don't need it, but we should add a try/catch and some basic error handling here.

KCCPMG (changes requested):
Good approach to getting all of the form data put into state. There are a few things I noticed here that I think should probably be changed (sorry for pointing them out now instead of earlier), and I added a few suggestions for how to deal with the button logic. We also still need to handle the question regarding frequency; I was thinking about doing a small PR that addressed that as well as renaming the "subgoal" table and procedures to reflect the term "benchmark". I might wait until Tuesday's meeting to bring it up with everybody, but let me know if you have any thoughts there, and also let me know if there's anything else I can do to help.

KCCPMG (approved):
Looks like everything works! I noticed a linter error: "61:6  warning  React Hook useEffect has missing dependencies: 'checkNumberFields' and 'checkTextFields'. Either include them or remove the dependency array". I don't need this fixed to approve, this has already been a ton of work. I'd been seeing a similar error on useEffect calls on my own project so I took a look into it, and I guess that the best practice is to declare functions inside your useEffect callback. I would have guessed that doing that would slightly hamper performance since I would think it would be constantly redeclaring the function, but what do I know.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -13,23 +36,45 @@ const CreateBenchmarkPage = () => {
...
+      materials: undefined,
+      instructions: undefined,
+      baseline_level: undefined,
+      target_level: undefined,
+      attempts_per_trial: undefined,
+      number_of_trials: undefined,
+    });
+
+  const steps = ["Instructional Guidelines", "Data Collection Guidelines"];
+
```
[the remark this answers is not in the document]
Comment (BrettEastman) [reply src/pages/students/[student_id]/goals/[goal_id]/create.tsx:87]: Great suggestion. I used the functions you wrote here. Instead of useEffect to disable the button, I opted for checks on click of the "Next" button and the submit button, so the user cannot move to the next page without filling out the fields in page one.

PR #353, opened 2024-07-17
change `npm i --ci` to `npm ci` to help ensure reproducible builds

nickvisut:
Best practice to use `npm ci` for reproducible builds (guarantees usage of our `package-lock.json`, so using it instead of `npm i --ci` in our Github Action. Looks like `--ci` may not exist as a switch either.


BrettEastman (approved):
This branch is out of date and needs to be updated to the current main. Otherwise, seems good. Would be helpful to have a better PR description, letting reviewers know exactly what this does, what the difference is, or if it is just a fix to remove the '--'

PR #354, opened 2024-07-17
Create dependabot.yml

nickvisut:
configure dependabot for npm and Github Actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+
+updates:
```
katconnors, line 8: Just to check, updates here is referring to a dependency check, and that's why the repo is listed as npm?

katconnors (commented):
Other than a quick clarification, the code looks good to me.

.github/dependabot.yml
nickvisut: that's right, it's a dep check. it's checking npm, but also GitHub Actions, for updates.

katconnors (approved):

PR #356, opened 2024-07-23
added check and test for IEP end dates

j-h-li:
Added check to IEP table schema in initial_migrations to make sure end date is greater than or equal to start date. created test in student.test.ts to check that having a smaller end date will throw an error when creating a new IEP or editing an IEP.


KCCPMG (approved):
Looks great! Good tests.

PR #382, opened 2024-08-14
Fix #369: Fixed Javascript console warnings on Create Benchmark page

### LINKED ISSUE #369: Fix stepper warning
When opening the Create benchmark page, a warning appears in the JavaScript console due to the stepper snippet. This ought to be debugged/addressed.

mrabbitt:
I found saw warnings when loading the "Create Benchmark" page, captured in this file:

[369_browser_console_warnings.log](https://github.com/user-attachments/files/16606471/369_browser_console_warnings.log)

These three appear to be caused by the Stepper expecting the icon components exp:
```
hydration-error-info.js:63 Warning: Received `false` for a non-boolean attribute `completed`.
hydration-error-info.js:63 Warning: Received `true` for a non-boolean attribute `active`.
hydration-error-info.js:63 Warning: Received `false` for a non-boolean attribute `error`.
```

This warning appears to be caused by the value property of a `TextArea` defaulting to `null`:
```
hydration-error-info.js:63 Warning: `value` prop on `textarea` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

### DIFF
--- a/src/pages/students/[student_id]/goals/[goal_id]/create.tsx
+++ b/src/pages/students/[student_id]/goals/[goal_id]/create.tsx
@@ -7,6 +7,7 @@ import {
   Box,
   Stack,
   Step,
+  StepIconProps,
   StepLabel,
   Stepper,
   TextField,
@@ -27,11 +28,21 @@ interface BenchmarkFormEntry {
   [key: string]: string | number | "";
 }
 
+const BenchmarkStepperIcon = (stepIconProps: StepIconProps) => {
+  const { completed = false } = stepIconProps;
+
+  if (completed) {
+    return <CheckCircle />;
+  } else {
+    return <TripOriginRounded />;
+  }
+};
+
 const CreateBenchmarkPage = () => {
   const router = useRouter();
   const { data: goal } = trpc.iep.getGoal.useQuery(
     { goal_id: router.query.goal_id as string },
-    { enabled: Boolean(router.query.goal_id) }
+    { enabled: Boolean(router.query.goal_id) },
   );
 
   const addSubgoalMutation = trpc.iep.addSubgoal.useMutation();
@@ -110,7 +121,7 @@ const CreateBenchmarkPage = () => {
       await router.push(
         `/students/${router.query.student_id as string}/goals/${
           router.query.goal_id as string
-        }`
+        }`,
       );
     } catch (error) {
       console.error("Error creating benchmark", error);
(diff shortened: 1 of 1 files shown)


src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -27,11 +28,21 @@ interface BenchmarkFormEntry {
   [key: string]: string | number | "";
 }
 
+const BenchmarkStepperIcon = (stepIconProps: StepIconProps) => {
```
Comment (mrabbitt) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:31]: This new component takes the same properties as `StepIcon`, so it can be safely passed as the `StepIconComponent` of a `StepLabel` without generating warnings.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -27,11 +28,21 @@ interface BenchmarkFormEntry {
...
+  const { completed = false } = stepIconProps;
+
+  if (completed) {
+    return <CheckCircle />;
+  } else {
+    return <TripOriginRounded />;
+  }
```
Comment (mrabbitt) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:38]: Since the Stepper will set `completed` appropriately whether or not the step for the current icon has been completed, we can keep track of which icon to use here.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -176,11 +187,7 @@ const CreateBenchmarkPage = () => {
           multiline
           rows={4}
           name={field.name}
-          value={
-            benchmarkFormState[field.name] !== ""
-              ? benchmarkFormState[field.name]
-              : null
-          }
```
Comment (mrabbitt) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:183]: As the fourth warning message states, the value shouldn't be null. I think the intent here should be to set the value to undefined or empty string when `benchmarkFormState[field.name]` is null or undefined. If the value is an empty string, it's fine to use it as is.

src/pages/students/[student_id]/goals/[goal_id]/create.tsx
```diff
@@ -263,11 +264,7 @@ const CreateBenchmarkPage = () => {
                       required
                       type="number"
                       name="baseline_level"
-                      value={
-                        benchmarkFormState["baseline_level"] !== ""
-                          ? benchmarkFormState["baseline_level"]
-                          : ""
-                      }
```
Comment (mrabbitt) [inline src/pages/students/[student_id]/goals/[goal_id]/create.tsx:270]: While this expression for `value` doesn't generate any warnings, I don't think it's actually doing what's intended. It's effectively the same as `value={benchmarkFormState["baseline_level"]}` without any ternary operator logic. I replaced this (and other similar instances) with a simpler expression that will use an empty string if `benchmarkFormState["baseline_level"]` is falsy (null or undefined).

I'm happy to revert this change if there's a reason for the original code that I missed.

mrabbitt:
See #385 instead.

PR #347, opened 2024-06-24
Created benchmark view new design

chengtchris1:
Address issue #341.
https://github.com/sfbrigade/compass/issues/341

PR #395, opened 2024-08-27
Created benchmark view new design

chengtchris1:


nickvisut (approved):
lgtm. seems identical to #347 aside from the four files discussed.

mrabbitt (approved):
I didn't check everything, but I confirmed at least that my changes from #385 are preserved. ✅

PR #399, opened 2024-08-30
Adds  test that ensures all api routes are protected

thomhickey:
To verify:

- pull code
- run tests
- -> they pass

### DIFF
--- a/src/backend/routers/public.ts
+++ b/src/backend/routers/public.ts
@@ -1,7 +1,7 @@
 import { publicProcedure, router } from "../trpc";
 
 export const publicRouter = router({
-  getFoo: publicProcedure.query(() => {
-    return "bar";
+  healthCheck: publicProcedure.query(() => {
+    return "Ok";
   }),
 });
--- a/src/backend/routers/public.ts
+++ b/src/backend/routers/public.ts
@@ -1,0 +1,7 @@
+import { publicProcedure, router } from "../trpc";
+
+export const publicRouter = router({
+  getFoo: publicProcedure.query(() => {
+    return "bar";
+  }),
+});
--- a/src/backend/routers/routes.test.ts
+++ b/src/backend/routers/routes.test.ts
@@ -29,8 +29,8 @@ test("All API endpoints are auth guarded", async (t) => {
   // create a mock response object that will be passed to the context
   const mockRes = {
     getHeader: () => undefined,
-    setCookie: () => void 0,
-    setHeader: () => void 0,
+    setCookie: () => undefined,
+    setHeader: () => undefined,
   } as unknown as NextApiResponse;
 
   // Create a mock context with no authentication
--- a/src/backend/routers/routes.test.ts
+++ b/src/backend/routers/routes.test.ts
@@ -1,0 +1,56 @@
+import test from "ava";
+import { trpcRouter } from "@/backend/routers/_app";
+import { createContext } from "@/backend/context";
+import type { NextApiRequest, NextApiResponse } from "next";
+import { TRPCError } from "@trpc/server";
+
+test("All API endpoints are auth guarded", async (t) => {
+  // Create a more complete mock of req and res
+  const mockReq = {
+    headers: {},
+    cookies: {},
+  } as unknown as NextApiRequest;
+
+  const mockRes = {
+    getHeader: () => undefined,
+    setCookie: () => void 0,
+    setHeader: () => void 0,
+  } as unknown as NextApiResponse;
+
+  // Create a mock context with no authentication
+  const ctx = await createContext({
+    req: mockReq,
+    res: mockRes,
+  });
+
+  // Create a caller with this context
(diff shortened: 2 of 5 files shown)


src/backend/routers/routes.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  // Create a more complete mock of req and res
+  const mockReq = {
+    headers: {},
+    cookies: {},
+  } as unknown as NextApiRequest;
+
+  const mockRes = {
+    getHeader: () => undefined,
+    setCookie: () => void 0,
+    setHeader: () => void 0,
```
KCCPMG, line 17: I'm curious, what's the reason that these are defined as returning void 0 instead of undefined?

src/backend/routers/routes.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (excludeList.includes(fullProcedureName)) {
+      continue;
+    }
+    const [routerName, procedureName] = fullProcedureName.split(".");
+
+    try {
+      await (
+        caller[routerName as keyof typeof caller] as Record<
+          string,
+          (...args: unknown[]) => Promise<unknown>
```
KCCPMG, line 44: I'm curious if there's a reason to use '...args: unknown[]' instead of just leaving the parameters empty?

KCCPMG (approved):
This looks ok to me, but I have to admit that I was in a bit over my head, so it could still be good to get some more eyes on it. It took me a while to understand what was going on, but on the bright side it inspired me to get my node debugger set up with an ava configuration. Some of this tRPC stuff is a bit more esoteric than than the bulk of what we're usually working with, so a few comments here and there (especially within the try/catch block for example) would probably be helpful to those of us who can review.

src/backend/routers/routes.test.ts
thomhickey: @KCCPMG typescript was complaining ... some esoteric type safety problem, I forget the details but initially it was returning undefined for all here.

src/backend/routers/routes.test.ts
thomhickey: @KCCPMG good catch! simplified in ebe1f54ae67a3657566bda37ea54b70a649f8bb8

thomhickey:
> This looks ok to me, but I have to admit that I was in a bit over my head, so it could still be good to get some more eyes on it. It took me a while to understand what was going on, but on the bright side it inspired me to get my node debugger set up with an ava configuration. Some of this tRPC stuff is a bit more esoteric than than the bulk of what we're usually working with, so a few comments here and there (especially within the try/catch block for example) would probably be helpful to those of us who can review.

@KCCPMG I added comments, although to be honest a long time ago I stopped worrying about my test code, I don't try to optimize it, I just make sure it tests the code that I actually care about. So you'll see me do a lot of brute force stuff with dummy data, etc, in my test code. Anyway, see the new comments maybe it will make it more clear.

src/backend/routers/routes.test.ts
KCCPMG: Hm ok. I changed it to undefined on my end when I was playing around with it and didn't get any type issues, maybe differences in our TS extensions?

thomhickey: you're totally right, changed to undefined ... I was having Cursor's AI fix linting/etc problems, it must have went overboard.

src/backend/routers/public.ts
```diff
@@ -0,0 +1,7 @@
+import { publicProcedure, router } from "../trpc";
+
+export const publicRouter = router({
+  healthCheck: publicProcedure.query(() => {
+    return "Ok";
+  }),
```
mrabbitt, line 6: To clarify: This is a placeholder for a health check, right? It doesn't actually check anything. I'm able to hit it with `http://localhost:3000/api/trpc/public.healthCheck` on my local, and it returns an successful response even if I've stopped the database and minio container.

It might be worth adding a comment to note that this is just a placeholder. Or maybe it's not needed as it's obvious to everyone who's not new to trpc like me. 😄

src/backend/routers/routes.test.ts
mrabbitt, line 1: Thanks for the thorough comments. 👍

mrabbitt (approved):
Looks good to me. I'm pretty new to trpc, so I might be missing something here, but it seems pretty straightforward.

Thanks, @thomhickey !

src/backend/routers/public.ts
thomhickey: @mrabbitt yes, this is just to verify the api is up and running to it at least checks that we are receiving and responding to api calls so I don't consider it just a placeholder. great callout to have it do more, though. I thought about a quick db call, but hadn't considered minio ... although we don't want it to take too long as this would be an endpoint we would configure with hosting providers that might alert us when the site is down, hence the need for it to be a public route. but most of all I wanted to introduce a public api route so that the excludeList in the test would be put to use and could be verified for correctness. thanks!

PR #404, opened 2024-09-04
Don't add 200px left margin for toolbar on sign-in page.

### LINKED ISSUE #403: Sign-in page layout not centered in wide desktop layout
When the sign-in page is viewed with with a browser window width of 992px or wider, the layout is not centered. It is shifted to the right.

It appears that a left margin of 200px is being added, probably to leave room for the navigation side bar which is displayed on most other pages, but not on the sign-in page.

Screen capture:

https://github.com/user-attachments/assets/8464c9a5-c1b6-4038-a9cc-1960e77b95fe

mrabbitt:
Fix to only add the `margin-left: 200px` needed for the navigation side bar on pages other than the sign-in page.

Fixes #403.

### DIFF
--- a/src/components/layout/Layout.tsx
+++ b/src/components/layout/Layout.tsx
@@ -3,6 +3,7 @@ import NavBar from "../navbar/NavBar";
 import $layout from "./Layout.module.css";
 import { useRouter } from "next/router";
 import { requiresLogin } from "@/client/lib/authenticated-page";
+import { useSession } from "next-auth/react";
 
 interface LayoutProps {
   children: React.ReactNode;
@@ -10,11 +11,13 @@ interface LayoutProps {
 
 const Layout = ({ children }: LayoutProps) => {
   const router = useRouter();
+  const { status } = useSession();
   // TODO?: Are there better ways to achieve this?
   //Sets purple background for student profile and instructions routes
   const isPurpleBg =
     router.asPath.includes("/studentprofile") ||
     router.asPath.includes("/instructions");
+  const hasNavBar = "authenticated" == status;
 
   return (
     <div className={$layout.container}>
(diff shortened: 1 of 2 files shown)


src/components/layout/Layout.tsx
```diff
@@ -3,18 +3,21 @@ import NavBar from "../navbar/NavBar";
...
 import { requiresLogin } from "@/client/lib/authenticated-page";
+import { useSession } from "next-auth/react";
 
 interface LayoutProps {
   children: React.ReactNode;
 }
 
 const Layout = ({ children }: LayoutProps) => {
   const router = useRouter();
+  const { status } = useSession();
```
mrabbitt, line 14: This replicates here the logic in the NavBar component to decide if the NavBar should be displayed:

https://github.com/sfbrigade/compass/blob/619dac8399708b8939479c3b126e9baae004a45f/src/components/navbar/NavBar.tsx#L109

src/components/layout/Layout.tsx
```diff
@@ -3,18 +3,21 @@ import NavBar from "../navbar/NavBar";
...
   // TODO?: Are there better ways to achieve this?
   //Sets purple background for student profile and instructions routes
   const isPurpleBg =
     router.asPath.includes("/studentprofile") ||
     router.asPath.includes("/instructions");
+  const hasNavBar = "authenticated" == status;
```
mrabbitt, line 20: This replicates here the logic in the NavBar component to decide if the NavBar should be displayed:

https://github.com/sfbrigade/compass/blob/619dac8399708b8939479c3b126e9baae004a45f/src/components/navbar/NavBar.tsx#L109

KCCPMG (approved):
Does what it says, looks good to me. I'm wondering if there's an original sin with this whole issue in that the Drawer component that the Navbar uses should be able to take care of the spacing itself without having to add a margin to its sibling, but that's a more complicated issue that someone else could look into later. In the meantime, I'm happy to approve this and to get all my future sessions start off on a better foot.

mrabbitt:
Thanks, @KCCPMG !

> I'm wondering if there's an original sin with this whole issue in that the Drawer component that the Navbar uses should be able to take care of the spacing itself without having to add a margin to its sibling, but that's a more complicated issue that someone else could look into later.

Agreed.

It might also be worth looking into [using per-page layouts](https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts#per-page-layouts) to have different layouts for different pages. Right now, it seems like the sign-in page is the only exception, so that's probably overkill. If in the future there's a need for more pages that either don't have the nav bar or require other layouts, it may be worth it.

For the moment though, the change in this PR seemed to me to be the least disruptive improvement. I'll merge it.

PR #405, opened 2024-09-05
Design systems search

### LINKED ISSUE #274: [Design System Components] Search
Please use the Figma file -> Design Systems -> Search for reference.

Try to standardize all inputs in the app to be in alignment with the Design Systems guidelines for primary, secondary, etc.

Use your best judgment

BeeSeeWhy:
Closes #274 

Seach Bar component created  

Component Example: 
Screenshot 2024-09-05 at 1 13 14 PM

Hover Example: 
Screenshot 2024-09-05 at 1 13 52 PM

Focused Example: 
Screenshot 2024-09-05 at 1 14 00 PM

Component In Action: 

https://github.com/user-attachments/assets/8e181d1f-8367-4c05-98f7-7c7bdab2d1ea


src/components/table/table.tsx
```diff
@@ -22,6 +21,8 @@ import { SelectableForTable } from "zapatos/schema";
 import emptyState from "../../public/img/empty-state.png";
 import Container from "@mui/material/Container";
 import Image from "next/image";
+import { SearchBar } from "../design_system/searchBar/SearchBar";
```
thomhickey, line 24: curious the relative path imports here? i would tend more toward @components/design_system/searchBar/SearchBar, which then makes table.tsx moveable.

src/components/table/table.tsx
```diff
@@ -153,9 +154,32 @@ function EnhancedTableToolbar({
...
+            <div
+              style={{
+                display: "flex",
+                justifyContent: "space-around",
+                alignItems: "center",
+              }}
+            >
+              <SearchBar
+                id="search-input"
+                placeholder="Search"
```
thomhickey, line 167: I think 'Filter' is more applicable to this UX. I know the figma design calls for 'Search', but this is filtering a list, not doing a traditional search, which would return a new list of search results. Just a comment, no big deal.

src/styles/Input.module.css
```diff
@@ -1,7 +1 @@
-/* TODO: Might use MUI for all inputs for consistency */
-.default {
-  border-radius: 2px;
-  border: 1px solid var(--outline);
-  margin: 2px 4px;
-  padding: 4px 2px;
-}
+/* moved to Form.module.css since this was not styling input, but was styling the form */
```
thomhickey, line 1: maybe remove this comment? it only makes sense if you know what used to be here.

thomhickey (commented):
Nice work, filters great! Had a few comments and then on Mobile at iPhone 14 Pro Max, the flow kind of falls apart, it's too crowded and I would recommend we style it down in height/font-size for mobile, but maybe the designs didn't call for that. 
image

src/components/table/table.tsx
BeeSeeWhy: I can definitely change that...I found that usage in another part of the app...but the @ makes more sense

src/components/table/table.tsx
BeeSeeWhy: I didn't build the search functionality, just the UI component. Do you think it would benefit us more to build a traditional search rather than the filtering?

src/styles/Input.module.css
BeeSeeWhy: Will do. Thanks.

src/components/table/table.tsx
thomhickey: No, I was just saying we should change the placeholder text from 'Search' to 'Filter', but I'm aware that you didn't create that placeholder value in this PR.

BeeSeeWhy: Do you think my listing error has to do with my version of node I’m using? It keeps saying prettier is not run, but I’ve even run it separate from the push check and says it’s ok

mrabbitt (changes requested):
@BeeSeeWhy  @nickvisut , @chengtchris1 , and I have run into this before where the behavior of prettier on local doesn’t match the GitHub check. What fixed it for us was: Remove`node_modules` and re-run`npm install`. (`npm ci` would probably work too.) After that, re-run prettier, and it will undo some changes  that it introduced incorrectly (mainly trailing commas).

Also, as noted in another comment I think you need to restore the lines you removed from the prettier ignore file.

The styling of the Search component looks ok to me on Desktop. Mobile could be better, as @thomhickey noted.

.prettierignore
```diff
@@ -1,5 +1,3 @@
-src/backend/db/zapatos
-.next
-.nsm
-.terraform
-*.css.d.ts
```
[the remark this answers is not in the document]
BeeSeeWhy: Ugh...this what I get for trying to work after getting a COVID vax  ... I'll try working on the mobile search as well

src/backend/db/zapatos/schema.d.ts
```diff
@@ -7,13 +7,13 @@ Copyright (C) 2020 - 2022 George MacKerron
 Released under the MIT licence: see LICENCE file
 */
 
-declare module 'zapatos/schema' {
-
-  import type * as db from 'zapatos/db';
+declare module "zapatos/schema" {
```
nickvisut, line 10: hmm looks like prettier was accidentally run on this file

src/components/design_system/input/Input.tsx
```diff
@@ -1 +1,30 @@
-export {};
+import React from "react";
```
nickvisut, line 1: seems like this is unneeded

src/components/design_system/input/Input.tsx
```diff
@@ -1 +1,30 @@
...
+import TextField, { TextFieldProps } from "@mui/material/TextField";
+import { styled } from "@mui/material/styles";
+
+export const TextInput = styled(TextField)<TextFieldProps>(({ theme }) => ({
```
nickvisut, line 5: can either use `theme` colors from `theme.ts` instead of `var(...)` below or remove it altogether

src/components/design_system/searchBar/SearchBar.tsx
```diff
@@ -1,9 +1,51 @@
...
- * 3) Search and find all use cases for your component (likely linting will tell you where they are) and update the import paths
- * 4) Check code for errors and delete this comment
- * 5) Push code to branch and do a PR referencing the specific issue task you took for issue # 255.
- * NOTE: If you want a css.d.ts file to be generated or updated to help with any type issues, you can run `npm run type-css`
- * */
-export {};
+import React from "react";
+import TextField, { TextFieldProps } from "@mui/material/TextField";
+import { styled } from "@mui/material/styles";
+import { Input, InputBase, InputBaseProps } from "@mui/material";
```
nickvisut, line 4: looks like some of these imports are extraneous and can be removed

nickvisut (changes requested):
left a few comments wrt lint errors and step theming, as well as a Q about prettier

src/backend/db/zapatos/schema.d.ts
nickvisut: (could you revert?)

francisli:
@nickvisut the implementation has changed, so I've resolved all of the prior comments...

Can you re-review...?

francisli:
Some screenshots of new implementation... also implements the X clear button added when there is input...

Screenshot 2025-04-23 at 11 46 13 AM
Screenshot 2025-04-23 at 11 46 22 AM

Issue #368, opened 2024-08-01
Update eng tickets


nickvisut:

nickvisut:
added "high priority" label for now; can rename later if need be

PR #409, opened 2024-09-12
add note to README about assigning labels for new issues

nickvisut:
Addresses #368

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,15 @@ Useful resources:
 - [Notion wiki](https://www.notion.so/sfbrigade/Compass-source-of-truth-db118a6d51344dccadc49d1988662d66)
 - [Slack channel](https://sfbrigade.slack.com/archives/C01R8E75N1K)
 
+## Filing a new issue
+
+If you run into an issue or would like to request a feature, please file an issue.
+
+When creating an issue, please assign the following issue labels where appropriate:
+
+- "design" label for a design or UX issue
+- "engineering" label for an engineering issue
+
 ## Contributing to our code base
 
 ### One time setup


README.md
```diff
@@ -7,6 +7,15 @@ Useful resources:
 - [Notion wiki](https://www.notion.so/sfbrigade/Compass-source-of-truth-db118a6d51344dccadc49d1988662d66)
 - [Slack channel](https://sfbrigade.slack.com/archives/C01R8E75N1K)
 
+## Filing a new issue
+
+If you run into an issue or would like to request a feature, please file an issue.
```
mrabbitt, line 12: For an issue that requests a new feature, it may require both UX design work and engineering work. What label(s) are appropriate in that situation?

mrabbitt (commented):
Looks fine to me, though perhaps it could use some more explanation regarding feature requests (see my other comment).

Also, will these instructions need to be updated again if we introduce issue templates which automatically add the labels (as per one of the line items in #368)?

This should probably be reviewed by someone else who's been around longer than me and is familiar with the way the team works.

README.md
nickvisut: it sounds like we may no longer need these labels. if so, we could get rid of them and the prefixes (at least moving forward)

@tam-janice, given the new board structure, will we no longer have separate eng and dev tickets? asking b/c traditionally the tickets were either prefixed with [ENG] or [UX] (or alternatively assigned similar labels).

nickvisut: we could also repurpose the labels for "engineering"-only (tech stuff, chores) and "UX"-only tickets (design that isn't meant to be handed off to dev), which would distinguish them from a story/task that goes through the whole lifecycle.

tam-janice: Here's what I'm thinking:

- We remove the prefixes [eng] and [UX]
- We will keep the column 'in design' 
- If something needs design, it'll jsut be assigned to me and I'll take care of it
- there will never be a ticket that reqires both UX and eng. If it's an eng ticket design should already be done

Does this work?

PR #412, opened 2024-09-15
fix: cleanup errors and warnings

thomhickey:
several warnings and errors were thrown in the console during student edit if the student has no iep data

### DIFF
--- a/src/components/iep/Iep.tsx
+++ b/src/components/iep/Iep.tsx
@@ -159,6 +159,7 @@ const Iep = ({ iep_id }: IepProps) => {
               src={noGoals}
               alt="no goals image"
               className={$Image.fitContent}
+              priority={true}
             />
             <p className={$Iep.noGoalText}>No goals yet</p>
             <p className={$Iep.noGoalTextSmall}>
--- a/src/pages/students/[student_id].tsx
+++ b/src/pages/students/[student_id].tsx
@@ -86,7 +86,7 @@ const ViewStudentPage = () => {
       grade: Number(data.get("grade")) || 0,
     });
 
-    if (startDate && endDate) {
+    if (activeIep) {
       editIepMutation.mutate({
         student_id: student.student_id,
         start_date: new Date(parseISO(data.get("start_date") as string)),
--- a/src/pages/students/[student_id].tsx
+++ b/src/pages/students/[student_id].tsx
@@ -86,11 +86,13 @@ const ViewStudentPage = () => {
       grade: Number(data.get("grade")) || 0,
     });
 
-    editIepMutation.mutate({
-      student_id: student.student_id,
-      start_date: new Date(parseISO(data.get("start_date") as string)),
-      end_date: new Date(parseISO(data.get("end_date") as string)),
-    });
+    if (startDate && endDate) {
+      editIepMutation.mutate({
+        student_id: student.student_id,
+        start_date: new Date(parseISO(data.get("start_date") as string)),
+        end_date: new Date(parseISO(data.get("end_date") as string)),
+      });
+    }
 
     handleMainState();
   };
@@ -163,7 +165,11 @@ const ViewStudentPage = () => {
             <p id="modal-modal-title" className={$CompassModal.editModalHeader}>
               Editing {student?.first_name || "Student"}&apos;s Profile
             </p>
-            <Typography id="modal-modal-description" sx={{ mt: 2 }}>
+            <Typography
+              id="modal-modal-description"
+              sx={{ mt: 2 }}
+              component="div"
+            >
               <Stack gap={0.5} sx={{ width: "100%" }}>
                 <form
                   className={$CompassModal.editForm}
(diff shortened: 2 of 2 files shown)


KCCPMG (approved):
Looks good to me. What's the reason for the `priority={true}` on the Images? There's arguably a larger structural issue with the database schema having student-to-iep as one-to-many and the logic of editing a student handled with the same logic than handles  editing an IEP, but that's probably a larger discussion and what you have here seems to work perfectly well for eliminating a few errors.

thomhickey:
> Looks good to me. What's the reason for the `priority={true}` on the Images? There's arguably a larger structural issue with the database schema having student-to-iep as one-to-many and the logic of editing a student handled with the same logic than handles editing an IEP, but that's probably a larger discussion and what you have here seems to work perfectly well for eliminating a few errors.

Yes @KCCPMG, codebase has larger issues that I didn't try to tackle, just wanted to get rid of the obvious warnings and errors for now. Happy to work on larger issues, maybe placing a UNIQUE constraint on iep.student_id so that only one row exists there per student. Or of course we could move the iep columns into the student table. re: the mutation code I'm also happy to go clean that up some more. Just feels like things are in flux in a way that makes me shy away from doing too much in the codebase right now.

re: priority={true} the warning was a next.js page-load optimization warning about having the largest (in kb) image on the page being below the fold. It was wrong, incidentally, the no-goals-icon.png image is above the fold. I suppose I might have looked to turn this check off in next config, but this was an easy fix. The warning only shows up in dev mode, doesn't show up in staging/prod. I just hate seeing any warnings/errors in the console when developing and in my experience they should always be squashed or they bite you later.

mrabbitt:
> re: priority={true} the warning was a next.js page-load optimization warning about having the largest (in kb) image on the page being below the fold. It was wrong, incidentally, the no-goals-icon.png image is above the fold. I suppose I might have looked to turn this check off in next config, but this was an easy fix. The warning only shows up in dev mode, doesn't show up in staging/prod.

@thomhickey I'm not seeing a console warning like this in my local dev environment when testing on the main branch. I'm running `npm run dev`, so I'm in dev mode. This should show up in the browser console, right? What's the text of the warning? Thanks!

thomhickey:
> > re: priority={true} the warning was a next.js page-load optimization warning about having the largest (in kb) image on the page being below the fold. It was wrong, incidentally, the no-goals-icon.png image is above the fold. I suppose I might have looked to turn this check off in next config, but this was an easy fix. The warning only shows up in dev mode, doesn't show up in staging/prod.
> 
> @thomhickey I'm not seeing a console warning like this in my local dev environment when testing on the main branch. I'm running `npm run dev`, so I'm in dev mode. This should show up in the browser console, right? What's the text of the warning? Thanks!

@mrabbitt here's what I see in my local with main branch for a student with no IEP:


https://github.com/user-attachments/assets/44418266-9fec-47f9-a750-6681fc8d52d5


The warning even comes up if the image is in full view when the page loads, at least for me it does. Here's what next says abvout this: https://nextjs.org/docs/api-reference/next/image#priority ... maybe I have a strict mode on that you don't or something?

mrabbitt (approved):
✅ Looks good to me. I was able to replicate several warnings showing up in the main branch which are all fixed in this PR's branch, and the changes make sense to me. Thanks, @thomhickey !

Thanks also for clarifying when the warning related to Image priority shows up. I was able to see it on my local.

For reference, here's all the warnings I noticed that were fixed:

* on loading the Student detail page without IEP:
	```
	console.js:288 Image with src "/_next/static/media/no-goals-icon.126be3a0.png" was detected as the Largest Contentful Paint (LCP). Please add the "priority" property if this image is above the fold.
	Read more: https://nextjs.org/docs/api-reference/next/image#priority
	overrideMethod @ console.js:288
	warnOnce @ warn-once.js:16
	eval @ get-img-props.js:346
	Show 3 more frames
	Show less
	```
* on clicking "Edit" on Student detail page without IEP:
	```
	hydration-error-info.js:63 Warning: validateDOMNesting(...): <fieldset> cannot appear as a descendant of <p>.
	hydration-error-info.js:63 Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
	hydration-error-info.js:63 Warning: validateDOMNesting(...): <form> cannot appear as a descendant of <p>.
	```
* on clicking "Save" button in edit student modal:
	```
	15:28:54.860 [student_id].tsx:89 TRPCClientError: [
	{
		"code": "invalid_type",
		"expected": "date",
		"received": "null",
		"path": [
		"start_date"
		],
		"message": "Expected date, received null"
	},
	{
		"code": "invalid_type",
		"expected": "date",
		"received": "null",
		"path": [
		"end_date"
		],
		"message": "Expected date, received null"
	}
	]
		at TRPCClientError.from (transformResult-6fb67924.mjs:9:20)
		at eval (httpBatchLink.mjs:197:101)
	console.error @ hydration-error-info.js:63
	window.console.error @ setup-hydration-warning.js:18
	overrideMethod @ console.js:288
	execute @ mutation.mjs:160
	await in execute
	mutate @ mutationObserver.mjs:89
	eval @ useMutation.mjs:27
	handleEditStudent @ [student_id].tsx:89
	...
	```

PR #413, opened 2024-09-17
Breadcrumbs refactor, testing using jest & storybook

mrabbitt:
### Included in this PR

* Refactored BreadcrumbsNav into two components for better separation of concerns and testability:
    * `Stateful BreadcrumbTrail`: Stateful container component that gets the URL path from and fetches student and para names via router calls. Passes data as props to `BreadcrumbTrail`.
    * `BreadcrumbTrail`: Stateless component which handles all layout/rendering using given props.
* Added basic "stories" files for BreadcrumbTrail which allows previewing and testing its appearance in various states using Storybook.
* Added support to compass project for testing using jest
    * Since ava is already used a test runner for the backend tests, separate scripts were added to package.json: 
        * `test:backend` runs existing tests for backend code with ava
        * `test:frontend` runs new frontend tests using jest
        * `test` runs both backend and frontend tests
* Added jest tests for Breadcrumbs:
    * `BreadcrumbTrail.test.tsx` tests BreadcrumbTrail component using [snapshot testing](https://jestjs.io/docs/snapshot-testing)
    * `BreadcrumbTrailUtils.test.ts` contains jest unit tests for a function used by BreadcrumbDesign to transform its input into breadcrumb segments.

### Not Included in this PR
* Fixes to breadcrumb behavior problems described in issue #396. I plan to address this in a later PR once some testing is in place.

### What input is especially requested from reviewers:
1. **Which test runner(s) do we want to use for compass?** ➡️ **Update: Based on feedback, we're going with option 1 (as implemented in this PR branch**
    * My thoughts: ava is a decent simple option for testing pure JavaScript/TypeScript code that has worked fine for backend tests. However, it's not as widely used and documented as jest, and I ran into significant problems trying to test components / JSX with it and was ultimately unsuccessful (will link to another draft PR with these attempts).  By comparison, the process getting jest to integrate with NextJS and test components was well documented and worked without issue.
    * Options:
        * **Option 1:** Use ava for backend and jest for frontend (what's currently implemented in this PR)
        * ~~Option 2: Use ava for all tests~~
            * This requires us to either drop the idea of unit testing or snapshot testing JSX code, or resolve the issues I ran into (link to ava draft PR)
            * (Note that the tests in `transformBreadcrumbs.test.ts` will work with ava without further research because they're testing pure TypeScript code without JSX.)
        * ~~Option 3: Use jest for all tests~~
[...]

PR #414, opened 2024-09-17
Breadcrumbs refactor, testing using ava & storybook

mrabbitt:
See the description of #413 for full background.

This PR is the same as #413 except that it attempts to use ava for component testing instead of jest.

See the "FIXME" comments in `Breadcrumbs.test.tsx` for open issues with component testing.

Note that tests in `transformBreadcrumbs.test.ts` should be working.


francisli:
Going to close this in favor of PR #505

Issue #437, opened 2024-10-11
Prettier intermittently fails in Github pipeline after running successfully locally


nickvisut:
We've historically had issues with prettier intermittently failing in the build pipeline for unknown reasons, even after running successfully on a local dev machine.

Only known workarounds (not guaranteed) are to either:
- nuke `node_modules` and run `npm run format` on the branch in question and push changes back up to Github
- install Node 18 and try same

This may be some kind of versioning issue. As per @francisli, it seems like we have two versions of Prettier being installed: 2 *AND* 3. That could be problematic, but we will have to research further to narrow it down.

We could also try to dissect and compare `package-lock.json` and the actual `node_modules` artifacts.

nickvisut:
cc @francisli

PR #438, opened 2024-10-11
Attempt to fix intermittent prettier issues

nickvisut:
We've historically had issues with prettier failing in the build pipeline for unknown reasons. This is likely some kind of versioning issue. Examining outputs of running v2.8.8 vs v.3.3.3 confirms different output, which can account for the issue as described in #437. However, we still don't know why the issue is intermittent nor why it sometimes seems to differ between environments. Forcing all versions of prettier to be v.3 or above and our code to conform to its defaults/our config is the goal of this PR.

This PR does the following:
- merges two prettier config files into one
- adds config property `trailingComma: "es5"` to preserve as much formatting as possible when going from v2 (where it's the default) and v3 (where `trailingComma: "all"` is the default instead)
- upgrades prettier from v2.8.8 to v.3.3.3 (latest)
- adds debugging statement in Github build pipeline to print out the prettier version for comparison between local and Github, which may be useful when there's an actual conflict in output (intermittent issue) and the build fails like in #437
- one-time formats a handful of files to address breaking changes in prettier v.3

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,1 +1,4 @@
-{}
+{
+  "tabWidth": 2,
+  "useTabs": false
+}
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,5 @@
 {
   "tabWidth": 2,
-  "useTabs": false
+  "useTabs": false,
+  "trailingComma": "es5"
 }
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "postinstall": "next build --no-lint && nsm build --skip-build --only-api-files && npm run type-css",
     "format": "prettier --write --ignore-unknown .",
     "format:check": "prettier --check --ignore-unknown .",
+    "format:version": "prettier --version",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "all:check": "npm run type:check && npm run type-css:check && npm run lint && npm run format:check && npm run test",
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "lint-staged": "^13.1.0",
     "nextjs-server-modules": "^4.7.0",
     "nodemailer-mock": "^2.0.1",
-    "prettier": "^2.8.8",
+    "prettier": "^3.3.3",
     "rewiremock": "^3.14.5",
     "storybook": "^8.1.11",
     "testcontainers": "^9.8.0",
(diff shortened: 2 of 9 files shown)


nickvisut:
Reading up on this, it sounds like this *may* have been a side effect of a newer variant of [this issue](storybookjs/storybook/issues/10339)

.prettierrc.json
```diff
@@ -1 +1,5 @@
-{}
+{
+  "tabWidth": 2,
+  "useTabs": false,
+  "trailingComma": "es5"
```
mrabbitt, line 4: I remember I briefly looked into this issue several weeks ago and found in the docs for [the trailingComma option](https://prettier.io/docs/en/options.html#trailing-commas): "Default value changed from es5 to all in v3.0.0," but I didn't figure out why the version of prettier that we were running was inconsistent. 👍

package.json
```diff
@@ -18,6 +18,7 @@
     "postinstall": "next build --no-lint && nsm build --skip-build --only-api-files && npm run type-css",
     "format": "prettier --write --ignore-unknown .",
     "format:check": "prettier --check --ignore-unknown .",
+    "format:version": "prettier --version",
```
mrabbitt, line 21: Suggestion:  Let's file an issue to remove this and the `prettier --version` in pr.yml in a few weeks if the problem doesn't recur.

mrabbitt: ...looks like this was discussed on slack already. 😃

mrabbitt (approved):
Looks good to me. ✅ 

Thanks, @nickvisut !

.prettierrc.json
nickvisut: yeah, what a weird bug. dep of a dep causing conflicts. just goes to show: Don't fully trust migration tools! I didn't expect prettier would be stealth-upgraded by another library. Will have to check eg `package-lock.json` more thoroughly next time. And will also have to study further how diff versions ran in diff envs at some point (maybe `prettier-fallback` from Storybook, unsure). 

also, it may make sense to define all of our prettier config explicitly instead of relying on defaults, although that will only help with breaking changes that alter config defaults.

nickvisut:
> Reading up on this, it sounds like this _may_ have been a side effect of a newer variant of [this issue](/storybookjs/storybook/issues/10339)

commenting here for add'l context ...

[Possibly related issue w/more details](https://github.com/storybookjs/storybook/issues/10192)

[Looks like Storybook made the dep on `prettier` less intertwined](https://github.com/storybookjs/storybook/issues/29084), so the newest version of Storybook may prevent this problem from happening again. We will have to see.

francisli (approved):
LGTM

Issue #450, opened 2024-10-16
UI should indicate and prevent duplicate task for same benchmark and para


canjalal:
### ACTUAL
If case manager attempts to assign a benchmark to Para "A", then tries to assign the benchmark to Para "A" once more, the UI will allow them to (checkbox is empty next to name).

~Furthermore, instead of showing an error, the app allows the case manager to actually create a duplicate task assigned to Para "A". This results in a duplicate row in the tasks table.~

As of finishing #439 (backend solution), we now see this error which does prevent duplicate task creation.
![image](https://github.com/user-attachments/assets/31f8b253-4f8c-4fe5-810d-3f0afcb9a2bb)

### EXPECTED
If case manager attempts to assign a benchmark to Para "A", then tries to assign the benchmark to Para "A" once more, the UI will show a checkbox next to their name b/c they are already assigned. The case manager can then choose to uncheck or recheck the checkbox to unassign or reassign.

Furthermore, if the front end UI checks are broken or bypassed, the app should not allow the case manager to actually create a duplicate task assigned to Para "A". Duplicate rows in the tasks table should not be created no matter what.

## Notes:

In the beginning, the fetching of a benchmark should also fetch all paras assigned to said benchmark through tasks

Unassigning a para will delete a task

trial_data should be associated with a benchmark_id rather than a task_id

canjalal:
Blocked by #439

canjalal:
Do we want to be able to unassign benchmarks? If we do this, it will destroy the data on the `due_date`,  `trial_count`, and the `seen` flag.

PR #454, opened 2024-10-18
Feat: add uniqueness constraint to task table

### LINKED ISSUE #439: Duplicate tasks for same benchmark and para should not be creatable (backend)
# ACTUAL

If case manager attempts to assign a benchmark to Para "A", then tries to assign the benchmark to Para "A" once more, the database will create a second task with the same benchmark and para

---

# EXPECTED

If case manager attempts to assign a benchmark to Para "A", then tries to assign the benchmark to Para "A" once more, the app should not allow the case manager to actually create a duplicate task assigned to Para "A". Duplicate rows in the `tasks` table should not be created no matter what. There should be a uniqueness constraint with the scope of benchmark and para.

canjalal:
Closes #439

Add uniqueness constraint on tasks to prevent assigning the same
benchmark (subgoal) to the same para (assignee).

What remains to be done is #450 , the frontend should indicate that a task has already been assigned to a given assignee and not allow the option to assign it to them again.

### DIFF
--- a/src/backend/db/zapatos/schema.d.ts
+++ b/src/backend/db/zapatos/schema.d.ts
@@ -2165,7 +2165,7 @@ declare module 'zapatos/schema' {
       */
       seen?: boolean | db.Parameter<boolean> | db.DefaultType | db.SQLFragment | db.SQLFragment<any, boolean | db.Parameter<boolean> | db.DefaultType | db.SQLFragment>;
     }
-    export type UniqueIndex = 'task_pkey' | 'uc_task';
+    export type UniqueIndex = 'subgoal_assignee_unique' | 'task_pkey';
     export type Column = keyof Selectable;
     export type OnlyCols<T extends readonly Column[]> = Pick<Selectable, T[number]>;
     export type SQLExpression = Table | db.ColumnNames<Updatable | (keyof Updatable)[]> | db.ColumnValues<Updatable> | Whereable | Column | db.ParentColumn | db.GenericSQLExpression;
--- a/src/backend/db/zapatos/schema.d.ts
+++ b/src/backend/db/zapatos/schema.d.ts
@@ -2165,7 +2165,7 @@ declare module 'zapatos/schema' {
       */
       seen?: boolean | db.Parameter<boolean> | db.DefaultType | db.SQLFragment | db.SQLFragment<any, boolean | db.Parameter<boolean> | db.DefaultType | db.SQLFragment>;
     }
-    export type UniqueIndex = 'task_pkey';
+    export type UniqueIndex = 'task_pkey' | 'uc_task';
     export type Column = keyof Selectable;
     export type OnlyCols<T extends readonly Column[]> = Pick<Selectable, T[number]>;
     export type SQLExpression = Table | db.ColumnNames<Updatable | (keyof Updatable)[]> | db.ColumnValues<Updatable> | Whereable | Column | db.ParentColumn | db.GenericSQLExpression;
--- a/src/backend/lib/db_helpers/case_manager.ts
+++ b/src/backend/lib/db_helpers/case_manager.ts
@@ -21,7 +21,7 @@ export async function createPara(
   from_email: string,
   to_email: string,
   env: Env
-): Promise<user.Selectable | undefined> {
+): Promise<user.Selectable> {
   const { first_name, last_name, email } = para;
 
   let paraData = await db
@@ -40,7 +40,7 @@ export async function createPara(
         role: "staff",
       })
       .returningAll()
-      .executeTakeFirst();
+      .executeTakeFirstOrThrow();
 
     // promise, will not interfere with returning paraData
     void sendInviteEmail(
--- a/src/backend/routers/case_manager.ts
+++ b/src/backend/routers/case_manager.ts
@@ -171,7 +171,7 @@ export const case_manager = router({
       );
 
       return await assignParaToCaseManager(
-        para?.user_id || "",
+        para.user_id,
         req.ctx.auth.userId,
         req.ctx.db
       );
--- a/src/backend/routers/iep.test.ts
+++ b/src/backend/routers/iep.test.ts
@@ -164,7 +164,13 @@ test("assignTaskToParas - no duplicate subgoal_id + para_id combo", async (t) =>
     authenticateAs: "case_manager",
   });
 
-  const para_id = seed.para.user_id;
+  const para_1 = seed.para;
+
+  const para_2 = await trpc.para.createPara.mutate({
+    first_name: "Foo",
+    last_name: "Bar",
+    email: "foo.bar@email.com",
+  });
 
   const iep = await trpc.student.addIep.mutate({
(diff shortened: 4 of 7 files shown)


src/backend/routers/iep.test.ts
```diff
@@ -51,7 +66,7 @@ test("basic flow - add/get goals, subgoals, tasks", async (t) => {
   const subgoal2Id = subgoal2!.subgoal_id;
 
   await trpc.iep.addTask.mutate({
-    subgoal_id: subgoal2Id,
```
canjalal, line 54: This subgoal_id is the same as the one for the assignTask on line 60 / 75. Therefore we had to add a new subgoal (subgoal1) for this to continue to work and not cause a uniqueness violation once the database migration is run

katconnors:
Quick q- to test the functionality of this, did you have to migrate again? I was trying to see if it worked on this branch and am still seeing the duplicate goals in Table Plus, so I assume I missed a step.

src/backend/db/zapatos/schema.d.ts
```diff
@@ -2165,7 +2165,7 @@ declare module 'zapatos/schema' {
       */
       seen?: boolean | db.Parameter<boolean> | db.DefaultType | db.SQLFragment | db.SQLFragment<any, boolean | db.Parameter<boolean> | db.DefaultType | db.SQLFragment>;
     }
-    export type UniqueIndex = 'task_pkey';
+    export type UniqueIndex = 'task_pkey' | 'uc_task';
```
katconnors, line 2168: I might suggest a different name than uc_task, for clarity
On that note, what is uc_task meant to be?

src/backend/routers/iep.test.ts
```diff
@@ -70,7 +85,7 @@ test("basic flow - add/get goals, subgoals, tasks", async (t) => {
   const gotSubgoals = await trpc.iep.getSubgoals.query({
     goal_id: goal1!.goal_id,
   });
-  t.is(gotSubgoals.length, 2);
+  t.is(gotSubgoals.length, 3);
```
katconnors, line 88: Could you please explain why this was updated? Thanks!

src/backend/routers/iep.test.ts
```diff
@@ -88,6 +103,62 @@ test("basic flow - add/get goals, subgoals, tasks", async (t) => {
...
+      subgoal_id: subgoal1Id,
+      assignee_id: para_id,
+      due_date: new Date("2024-03-31"),
+      trial_count: 1,
+    });
+  });
+
+  t.is(
+    error?.message,
+    'duplicate key value violates unique constraint "uc_task"'
```
katconnors, line 158: I'd suggest adding more detail to this error message.

katconnors (commented):
Looks good- just some minor edits and questions for my own understanding.

src/backend/db/zapatos/schema.d.ts
canjalal: "uniqueness constraint: task"

canjalal:
> Quick q- to test the functionality of this, did you have to migrate again? I was trying to see if it worked on this branch and am still seeing the duplicate goals in Table Plus, so I assume I missed a step.

I reset the entire database

src/backend/routers/iep.test.ts
canjalal: Previously, one of the subgoals was created, and used twice in the test to create a task with the same para. With the addition of the new database constraint, the creation of the second task failed. So I made a new subgoal in order for both tasks to be able to be created successfully, and thus the number of subgoals for that goal is now 3 instead of 2.

src/backend/routers/iep.test.ts
canjalal: Done!

src/backend/routers/iep.ts
```diff
@@ -135,6 +135,19 @@ export const iep = router({
...
+        .selectFrom("task")
+        .where("subgoal_id", "=", subgoal_id)
+        .where("assignee_id", "=", assignee_id)
+        .selectAll()
+        .executeTakeFirst();
+
+      if (existingTask) {
+        throw new Error(
+          "Task already exists: This subgoal has already been assigned to the same para"
+        );
```
canjalal, line 148: Throw error instead of making duplicate task. On the frontend we want to try to make sure this doesn't happen.

nickvisut:
lgtm but I was pairing, so need another reviewer to approve/merge

katconnors (approved):
Looks good to me, thanks for making those edits.

src/backend/lib/db_helpers/case_manager.ts
```diff
@@ -40,7 +40,7 @@ export async function createPara(
         role: "staff",
       })
       .returningAll()
-      .executeTakeFirst();
+      .executeTakeFirstOrThrow();
```
canjalal, line 43: We currently don't handle any cases where we test for it returning 'undefined', therefore I think having it throw and forcing us to deal with the error is a better option

src/backend/routers/case_manager.ts
```diff
@@ -171,7 +171,7 @@ export const case_manager = router({
       );
 
       return await assignParaToCaseManager(
-        para?.user_id || "",
+        para.user_id,
```
canjalal, line 174: The previous way doesn't make sense IMO

src/backend/routers/iep.test.ts
```diff
@@ -88,6 +103,124 @@ test("basic flow - add/get goals, subgoals, tasks", async (t) => {
...
+    });
+  });
+
+  t.is(
+    error?.message,
+    "Task already exists: This subgoal has already been assigned to the same para"
+  );
+});
+
+test("assignTaskToParas - no duplicate subgoal_id + para_id combo", async (t) => {
```
canjalal, line 162: We added this test because assignTaskToParas also creates a  new para record

src/backend/routers/iep.ts
```diff
@@ -160,6 +173,19 @@ export const iep = router({
...
       const { subgoal_id, para_ids, due_date, trial_count } = req.input;
 
+      const existingTasks = await req.ctx.db
+        .selectFrom("task")
+        .where("subgoal_id", "=", subgoal_id)
+        .where("assignee_id", "in", para_ids)
+        .selectAll()
+        .execute();
+
+      if (existingTasks.length > 0) {
```
canjalal, line 183: If any task for this benchmark already includes one or more of the selected para's, then throw the error

src/components/benchmarks/BenchmarkAssignmentModal.tsx
```diff
@@ -173,6 +183,12 @@ export const BenchmarkAssignmentModal = (
             />
           </Box>
         )}
+
+        {errorMessage && (
+          <Box className={$benchmark.benchmarkDescriptionBox}>
+            {errorMessage}
```
canjalal, line 189: Output the error here for now, in next PR we will aim to prevent this from happening

katconnors (approved):
Both the code and the feature functionality look good to me.

PR #455, opened 2024-10-18
Builds out roles + role-specific landing UX + tighten up auth guards

thomhickey:
- migration changes users.role to four roles: admin, case_manager, para, user
- user.role added to session context (this is done via db lookup on each request - can be optimized to instead store user.role in session at login that's a @todo)
- new UserType enum for role string constants, and applied everywhere instead of using string literals
- trpc.ts exports four procedures for trpc use and implements cascading access control user -> para -> case_manager -> admin. so para can do everything user can do, case_manager can do everything para and user can do, etc.
- all endpoints tightened up to use the appropriate trpc procedure according to role base access control
- oauth redirect route is now / instead of /students, with each role being forwarded to appropriate landing page including a new 'sorry' page that says you cannot access compass
- tests updated

to test:

- change your role in db
- log out/log in
--> user goes to /sorry 
--> para goes to /benchmarks and has limited nav menus
--> case_manager goes to /students and has appropriate nav menus
--> admin goes to /students but has all nav


src/backend/trpc.ts
```diff
@@ -1,6 +1,34 @@
...
+
+const ROLE_LEVELS: RoleLevel = {
+  user: 0,
+  para: 1,
+  case_manager: 2,
+  admin: 3,
+};
+
+// Function to compare roles
+function hasMinimumRole(auth: Auth, requiredRole: UserType): boolean {
```
nickvisut, line 22: this function assumes that each role is a superset of the prior one; does this mean a case manager should be able to do what a para does?

src/types/auth.ts
```diff
@@ -0,0 +1,27 @@
...
+  };
+}
+
+export interface CustomAdapterUser extends AdapterUser {
+  profile?: {
+    role: UserType;
+  };
+}
+
+export enum UserType {
```
nickvisut, line 22: nice!

nickvisut:
added Compass Engineering Reviewers as test of the new team

src/backend/trpc.ts
canjalal: Yes. Uncommon work flows could be removed from the UI, but I'm thinking in the future there could be an administrator panel that could change database records through trpc calls

thomhickey: Yes, @nickvisut for now cascading roles on the backend seems like a good fit for POC/MVP. And as mentioned by Vince, the front-end doesn't do much in the way of auth guards, it just hides ui, but entering a route by hand will render the page, even if the api won't fill the page with data. We can add stories for those use cases as well.

src/backend/routers/case_manager.test.ts
```diff
@@ -277,9 +306,28 @@ test("addStudent - invalid email", async (t) => {
...
 });
 
+test("addStudent - paras do not have access", async (t) => {
+  const { trpc } = await getTestServer(t, { authenticateAs: UserType.Para });
+
+  const error = await t.throwsAsync(async () => {
+    await trpc.case_manager.addStudent.mutate({
+      first_name: "Foo",
+      last_name: "Bar",
+      email: "invalid-email",
```
canjalal, line 316: Nitpick: this shouldn't be an invalid email because that's not what we're testing. I'm going to fix this

src/backend/tests/fixtures/get-test-server.ts
```diff
@@ -11,9 +11,10 @@ import ms from "ms";
 import builtNextJsFixture from "../../../../.nsm";
 import { getTestMinio } from "./get-test-minio";
 import superjson from "superjson";
+import { UserType } from "@/types/auth";
 
 export interface GetTestServerOptions {
-  authenticateAs?: "case_manager" | "para" | "admin";
+  authenticateAs?: UserType.CaseManager | UserType.Para | UserType.Admin;
```
canjalal, line 17: Do we want to just set the type to UserType, and make it default to UserType.User if it's not specified?

src/pages/sorry.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import $box from "@/styles/Box.module.css";
+import $typo from "@/styles/Typography.module.css";
+
+const SorryPage: React.FC = () => {
```
canjalal, line 5: If we aren't going to use this element in this PR, let's remove it and add it to a new branch split from this one.

src/types/auth.ts
canjalal: Honestly I'm wondering if it would be possible to just use the default TypeScript enum where UserType.User = 0, UserType.Para = 1, and then use it as our ROLE_LEVEL, and instead have some fixed function called getUserTypeString(userType: UserType) that returns "user", "para", "case_manager", "admin", and is used only for writing to the database, which stores these strings rather than the enum.

src/backend/routers/para.test.ts
```diff
@@ -199,3 +250,25 @@ test("getMyTasks", async (t) => {
...
+test("getMyTasks - paras do have access", async (t) => {
+  const { trpc } = await getTestServer(t, { authenticateAs: UserType.Para });
+
+  await t.notThrowsAsync(async () => {
+    await trpc.para.getMyTasks.query();
+  });
+});
+
+test("getMyTasks - regular users don't have access", async (t) => {
+  const { trpc } = await getTestServer(t, {});
```
canjalal, line 263: See my comment on making authenticateAs default to UserType.User if not specified, and have authenticateAs just have type UserType

src/backend/tests/fixtures/get-test-server.ts
canjalal: ^^ Just tried this, there's not much benefit because the seed file doesn't have a "UserType.User" key so it becomes more trouble than it's worth

francisli (approved):

PR #461, opened 2024-11-04
Update README.md to make clear to login before seeding

canjalal:
Make it more clear you need to log in with Google OAuth to generate the first user before you can run the seed script.


katconnors (approved):

nickvisut (approved):
helpful, thanks!

PR #462, opened 2024-11-07
Feature/add sort date to assigned

nickvisut:
Created At date added to task table to allow for sorting.


src/pages/benchmarks/index.tsx
```diff
@@ -11,22 +11,50 @@ import { useState } from "react";
...
+  const { data: tasksData, isLoading } = trpc.para.getMyTasks.useQuery();
 
   const handleTogglePara = () => {
     setIsPara(!isPara);
   };
 
+  const getSortedTasks = () => {
+    if (!tasksData) return [];
+
+    return [...tasksData].sort((a, b) => {
```
canjalal, line 31: Can we put all of this in a useEffect, rather than making a copy that gets sorted? I don't see a problem with changing the data that came in from the trpc

src/pages/benchmarks/index.tsx
```diff
@@ -142,6 +142,20 @@ function Benchmarks() {
               */
               }
 
+              {/* simple sort pill POC (see TODO above) */}
+              <button
```
canjalal, line 146: For a future PR: This should be a <PillButton> component with props for sortCriterion ("created_at") and title ("Sort by date")

src/pages/benchmarks/index.tsx
```diff
@@ -142,6 +142,20 @@ function Benchmarks() {
               */
               }
```
canjalal, line 143: ^^ Were you meant to delete this commented out code? Lines 125 - 143

francisli (approved):

canjalal (approved):
Looks good, nothing blocking

src/pages/benchmarks/index.tsx
nickvisut: updated with clarifying comment

canjalal (approved):

src/pages/benchmarks/index.tsx
nickvisut: cool, will dev in Storybook as discussed

PR #477, opened 2024-12-06
Adds storybook testing to ci pipeline

thomhickey:
refactored the Edit studend modal into a component, added a story for it and some basic tests, then added all the piping to get the tests to run on PR using github pr.yml.


francisli (approved):

.storybook/preview.tsx
```diff
@@ -8,7 +8,7 @@ import "../src/styles/globals.css";
 
 const preview: Preview = {
   parameters: {
-    actions: { argTypesRegex: "^on[A-Z].*" },
+    // actions: { argTypesRegex: "^on[A-Z].*" },
```
thomhickey, line 11: we should do something with this other than leave it commented out

francisli: If our stories/tests are working/passing as expected with it commented out, then I think we can just remove it. This global regexp-based configuration has been deprecated and needs to be configured in stories as needed:

https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#implicit-actions-can-not-be-used-during-rendering-for-example-in-the-play-function

PR #484, opened 2024-12-21
Fixes #468 (extra trial data) - update to prevent rerender

### LINKED ISSUE #468: Incorrect Trial/extra trial data instance creation
When a user navigates directly to a new benchmark (without trial data) via the url benchmarks/[id]/instructions url and then clicks "collect data", trial 2 is sometimes displayed instead of trial 1.

Further insight: I monitored the table "trial_data", and after clicking collect data (at the above url), two trial data instances are created.

See benchmark-instructions branch for some comments on behavior (https://github.com/sfbrigade/compass/commit/f289283b0c48620d809e655c012075866d412e50)

image
image

katconnors:
Behavior described in issue #468 is no longer occuring. 
Source of the problem was that useEffect was causing a rerender of the benchmark component in development.

### DIFF
--- a/src/pages/benchmarks/[benchmark_id]/index.tsx
+++ b/src/pages/benchmarks/[benchmark_id]/index.tsx
@@ -71,7 +71,7 @@ const BenchmarkPage = () => {
   const [currentTrialIdx, setCurrentTrialIdx] = useState(0);
   const currentTrial = task?.trials[currentTrialIdx] || null;
 
-  const trialAddedRef = useRef(false);
+  const trialAddedRef = useRef<boolean>(false);
 
   const hasInputChanged =
     currentTrial?.notes !== notesInputValue ||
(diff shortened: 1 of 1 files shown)


src/pages/benchmarks/[benchmark_id]/index.tsx
```diff
@@ -71,7 +71,7 @@ const BenchmarkPage = () => {
   const [currentTrialIdx, setCurrentTrialIdx] = useState(0);
   const currentTrial = task?.trials[currentTrialIdx] || null;
 
-  const [trialAdded, setTrialAdded] = useState(false);
+  const trialAddedRef = useRef(false);
```
Comment (canjalal) [inline src/pages/benchmarks/[benchmark_id]/index.tsx:74]: nitpick: can you type this as `useRef<boolean>(false)`?

canjalal (approved):

src/pages/benchmarks/[benchmark_id]/index.tsx
Comment (katconnors) [reply src/pages/benchmarks/[benchmark_id]/index.tsx:74]:
```suggestion
  const trialAddedRef = useRef<boolean>(false);
```

PR #488, opened 2025-01-24
[Closes #487] Feature/move trial data to benchmark

### LINKED ISSUE #487: Detangle Tasks and Benchmarks
- [ ] Move trial_data to belong to benchmarks rather than tasks
- [ ] getBenchmarkTrialData should return only benchmark, not tasks
- [ ] Make backend helper method that checks that current user is allowed to see / edit benchmark (including trial data)
- [ ] Frontend components should handle benchmark_id,  not task_id
- [ ] Tasks should be marked as 'seen' by backend, not frontend

francisli:


francisli:
I've fixed broken tests, updated the backend implementation for assignTasks, and updated the corresponding implementation in the assignment modal.

Next step: update the modal implementation to show the currently assigned paras when editing.

canjalal (approved):
Good work everyone!

PR #490, opened 2025-02-05
WIP: add build folder to separate dev server from build artifacts

nickvisut:
`.next` folder is used by both dev server and build process. this could be confusing as trying to both during a coding session could introduce unexpected issues.


nickvisut:
Hmm lint task is running on the build folder. Will need to remedy this.

nickvisut:
Closing this because it's not obvious that separate build folders for dev server and production build is widely supported (for example, on Vercel).

PR #496, opened 2025-03-07
Start creating generic Edit/Create BenchmarkForm

canjalal:
The aim of this PR is to allow the editing of benchmarks after creation. This will be done by separating out the current Create Benchmark Form into a BenchmarkForm component that can be used to either create or edit benchmarks, and then pointing the Edit Benchmark action (currently stubbed to "To be implemented") to point to an edit route that will call BenchmarkForm. The existing create route will also then call this new form.


francisli (approved):
This works well... per your suggestion, I've changed the routing so that the create/edit routes now use the path instead of a query param...

PR #508, opened 2025-06-03
Logo fix for HTML email template

katconnors:
Updated logos to PNG format.
Note that the default logo is being used in the reference because the white logo doesn't display the letter.


francisli (approved):
I've updated the PNG assets so that the white one renders properly with a transparent background and changed the template header to use it. I've also removed the unnecessary closing tags identified by Gemini which would have broken the template.

PR #532, opened 2025-10-02
Feature/chart view

MarianaES:
Include chart to visualize trials daily and their corresponding success rates. Considering highest and lowest scores. 
image

Multiple Trials Tooltip
image

Single Trial Tooltip
image

Range Tooltip (High & Low)
image


nickvisut:
Looks really good and quite clean!

More of a UX thing: I noticed that the exploded data points are no longer set. @francisli mentioned the range markers are meant to take their place? If so, it does make the graph a bit confusing, especially with the label "Points".

The tabs make senss, although I wonder if it might be better to add a "Averaged points" toggle to the "Points" tab:
- Averaged points ON: each date w/multiple trials collapsed into a multi-point
- Averaged points OFF: show *all* trials as their own solo points

This would allow the tabs to further simplify things by getting rid of "Single trials" vs "Multiple trials" altogether.

It looks like you were looking into that at some point with the `Chip` with label "Individual trials visible" (~= "Averaged points off").

src/pages/students/[student_id]/goals/[goal_id]/benchmarks/[benchmark_id]/view.tsx
```diff
@@ -103,19 +107,32 @@ const ViewBenchmarkPage: NextPageWithBreadcrumbs = () => {
   const soloPoints: SoloPoint[] = [];
   const bulkPoints: BulkPoint[] = [];
 
-  const getDateFromPtNumber = (
-    pointNumber: number,
-    pointArray: DatePoint[]
-  ) => {
-    return new Date(pointArray[pointNumber].x).toDateString();
-  };
+  const allTrials: Array<{
```
Comment (nickvisut) [inline src/pages/students/[student_id]/goals/[goal_id]/benchmarks/[benchmark_id]/view.tsx:110]: don't see this variable used anywhere

nickvisut (approved):
As in my prior comment, looks really good. I just added a couple of review comments for relatively minor things. You could probably merge as-is and fix in a follow-up PR with one exception.

The exception is regarding my UX feedback in the prior comment: Is `allPoints` being unused a bug? If not, then disregard, although then it'd be great to see a response to the UX feedback at some point.

src/components/benchmarks/CustomItemTooltip.tsx
```diff
@@ -4,24 +4,76 @@ import Popper from "@mui/material/Popper";
...
+      if (dateKey && datePoints[dateKey]) {
+        return datePoints[dateKey];
+      }
+    }
+
+    return [];
+  };
+
+  const trials = getTrialsForPoint();
+  const formattedDate = getFormattedDate();
```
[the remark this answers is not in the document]
MarianaES: Done, you are right it is better to be consistent.

src/pages/students/[student_id]/goals/[goal_id]/benchmarks/[benchmark_id]/view.tsx
Comment (nickvisut) [reply src/pages/students/[student_id]/goals/[goal_id]/benchmarks/[benchmark_id]/view.tsx:110]: ah oops, I'll unresolve this when you push the relevant commit (or GitHub will mark it "outdated")

nickvisut:
> @nickvisut Thank you for your feedback ◡̈ The reason behind using the range markers instead of showing all individual points are:
> 
>     * Keep the graph clean: When I was exploring having many trials per date, it may feel too crowded and hard to read. Range markers show the spread more clearly.
> 
>     * Trials within a date are independent (or at least that is my understanding please correct me if not): Meaning that the don't really connect to each other. Showing them all with a line as first we first thought may give a bad idea to the user.
> 
>     * Follow a little bit closer the original design: Not fully, because the dot line connecting the range markers was not an easy implementation on Material UI.
> 
> 
> However, I'm open to other ideas. We could:
> 
>     * Add the "averaged points" toggle you suggested and show individual points at first. (It would be nice to receive feedback from the professors on what metric is more useful, average or all the trials, or both. Even if the range markers are useful at all. 😅)

That makes sense! Hopefully, you all can get feedback to inform future changes. Hmm I even wonder if the educators might even prefer to see the trend over trials without them being pinned to dates, come to think of it. Who knows?

In any case, as you saw, I already approved your PR so that you could merge as soon as you made the smaller changes.

PR #541, opened 2025-11-07
Better type checking.

### LINKED ISSUE #516: Validation issue
- Student name (First name and Last name) accepts any char as input.
- The student grade accepts '0'

MapleSyrupy:
Teachers can no longer set grade to less than 1 or more than 13, and also cannot use special characters  (or number)  for names, unless its a dash.


MapleSyrupy:
Error handling is a bit of a mess with this PR, but I don't think that is the main focus of the issue, so it's more of a stopgap measure.

